PR #4, opened 2019-09-18
Sunstop

nritsche:
- Split SunClean into two tasks to allow saving intermediate data products
- edit config file for new sun tasks
- replace dependency on pyephem with skyfield
- Update SolarCalibration and SolarClean to handle extended solar emission


jrs65:
@wulfda02 is going to make a new branch with the useful solar analysis code.

PR #49, opened 2020-07-30
Add decorrelation correction to holography fringestop pipeline

tristpinsm:


tristpinsm:
should I get this merged?

PR #53, opened 2020-09-14
feat(daily): save a single freq ringmap

nritsche:
For daily validation.
If you want to add more frequencies, please add a code-change-suggestion-comment.


jrs65:
@nritsche do you still want this one to go in? I can merge it and we'd have it for the rev04 processing. I do recall that we added a second ringmap frequency so I think this at least needs updating with the new frequency.

nritsche:
Replaced by https://github.com/chime-experiment/bondia/pull/56

PR #61, opened 2020-12-23
Fix sphinx build and put it on readthedocs

nritsche:
https://chime-experiment.github.io/ch_pipeline


.github/workflows/sphinx.yml
```diff
@@ -0,0 +1,38 @@
...
+          mkdir docs
+          sphinx-build -b html doc/ docs
+          touch docs/.nojekyll
+      - name: Commit site
+        run: |
+          git config --global user.name 'Rick Nitsche (automated)'
+          git config --global user.email 'nritsche@users.noreply.github.com'
+          git add .
+          git commit -am "Automated sphinx build"
+          git push --force origin gh-pages
```
anjakefala, line 38: Why `--force`?

nritsche: Good question. Mostly because this is copypasta. I also questioned this, but concluded: "why not?" this is committing the compiled html docs to a special branch. If it builds, we always want to push it. Maybe there could be 2 instances of this running at the same time, and we don't want the CI to fail because of that (even though the docs wouldn't be complete after they ran both at the same time?). I could be convinced to remove it...

doc/dev.rst
```diff
@@ -0,0 +1,54 @@
...
+^^^^^^^^^^^^
+
+Dependencies should be installable python packages. This means that they must
+have a ``setup.py`` script in their root directory, and should be installable
+using `python setup.py install`. They are kept track of in a ``requirements.txt``
+file. This file can contain references to exact versions of dependencies, using
+both version tags, and commit hashes. An example ``requirements.txt`` file is
+given below::
+
+    -e git+https://github.com/radiocosmology/caput@ee1c55ea4cf8cb7857af2ef3adcb2439d876768d#egg=caput-master
```
anjakefala, line 32: I think we want to discourage installing in `editable mode`! Because then pip interprets the version as the commit hash.

I think we also want to discourage `#egg`, my understanding is that it is deprecated.

anjakefala (approved):

.github/workflows/sphinx.yml
anjakefala: That makes sense. I do not feel strongly about removing it.

doc/conf.backup.py
```diff
@@ -0,0 +1,301 @@
...
+
+# -- General configuration ------------------------------------------------
+
+# If your documentation needs a minimal Sphinx version, state it here.
+# needs_sphinx = '1.0'
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
```
anjakefala, line 33: It would be cool to have a comment for what each extension gives us, but not necessary!

requirements.txt
```diff
@@ -4,3 +4,4 @@ ch_util @ git+https://github.com/chime-experiment/ch_util.git
 cora @ git+https://github.com/radiocosmology/cora.git
 driftscan @ git+https://github.com/radiocosmology/driftscan.git
 draco @ git+https://github.com/radiocosmology/draco.git
+ephem
```
jrs65, line 7: In theory ephem should be deprecated and removed from all our code. I'd rather not have it sneaking back in. My plan was basically to remove it everywhere, and then if there is anything still using ephem it will break and get fixed to use Skyfield instead. Do you absolutely need it to here to build the docs? If so, can you rework things such that it can be removed from the requirements.

jrs65 (changes requested):

requirements.txt
nritsche: `ephem.Sun()` is used by `ch_pipeline.analysis.solar` only. Can I replace that with something? `caput.time.skyfield_wrapper.ephemeris["sun"]`?

jrs65: That's probably fine. I'm pretty sure that the new ephemeris changes have already broken that code anyway.

jrs65 (approved):
I'd forgotten about this. Thanks for getting all this doc stuff sorted! You're a hero.

PR #63, opened 2021-01-06
Quarterstack

jrs65:


.github/workflows/main.yml
```diff
@@ -44,12 +44,12 @@ jobs:
 
     - name: Install pip dependencies
       run: |
-        pip install git+https://github.com/chime-experiment/ch_util-stub.git
+        pip install --use-deprecated=legacy-resolver git+https://github.com/chime-experiment/ch_util-stub.git
```
anjakefala, line 47: =C

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+        output_name: "dspec.h5"
+"""
+
+
+class QuarterStackProcessing(base.ProcessingType):
+    """Stacks the daily data within quarters, subdivided into interleaved jack knifes.
+
+    This uses opinions in the dataflag database `chimedb.dataflag` to determine which
+    days are good and bad for each revision of the daily processing. It will then
+    take the each good day (taking from the latest revision if multiple revisions
```
nritsche, line 149: typo

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+        available good days into the individual stacks.
+        """
+
+        days = {}
+
+        core.connect()
+
+        # Go over each revision and construct the set of LSDs we should stack, and save
+        # the path to each.
+        # NOTE: later entries is `daily_revisions` will override LSDs found in earlier
```
nritsche, line 213: typo

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+                self.root_path
+                if self.default_params["daily_root"] is None
+                else self.default_params["daily_root"]
+            )
+            daily_rev = daily.DailyProcessing(rev, root_path=daily_path)
+
+            # Get all the bad days in this revision
+            revision = df.DataRevision.get(name=rev)
+            query = (
+                df.DataFlagOpinion.select(df.DataFlagOpinion.lsd)
```
nritsche, line 227: Should this not depend on data flags rather than opinions?
This only makes sense because at the moment there is a small number of opinions entered by people you trust to make good decisions.

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+            # Get all the good days
+            query = (
+                df.DataFlagOpinion.select(df.DataFlagOpinion.lsd)
+                .distinct()
+                .where(
+                    df.DataFlagOpinion.revision == revision,
+                    df.DataFlagOpinion.decision == "good",
+                )
+            )
+            good_days = [x[0] for x in query.tuples()]
```
nritsche, line 245: What's the rest of the tuple? Can this be done with a `.get()` or similar instead?

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+
+            for d in daily_rev.ls():
+
+                # Filter out known bad days here
+                if (int(d) in bad_days) or (int(d) not in good_days):
```
nritsche, line 250: `ProcessingType.ls()` looks like it returns a list of names (`str`) that it gets from file names. Should we catch `ValueError`s here when parsing them?

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+
+        lsd_partitions = {}
+
+        # For each quarter divide the LSDs it contains into a number of partitions to
+        # give jack knifes
+        for quarter in quarters:
+
+            lsds_in_quarter = sorted(np.array(lsds)[yq == quarter])
+
+            # Skip quarters with two few days in them
```
nritsche, line 277: typo

ch_pipeline/processing/quarterstack.py
```diff
@@ -0,0 +1,332 @@
...
+
+        This includes any that currently exist or are in the job queue.
+        """
+
+        return list(self._revparams["stacks"].keys())
+
+    def _finalise_jobparams(self, tag, jobparams):
+        """Modify the job parameters before the final config is generated.
+
+        Unfortunately this needs to by hand generate the list of daily file paths to
```
nritsche, line 303: I agree we need to change this.

ch_pipeline/processing/quarterstack.py
nritsche: This should be how you get the lsd out of dataflags that were created by voting: https://github.com/chime-experiment/chimedb_dataflag/blob/409b18d309ecf16bdec8f306094f84ee057895dc/chimedb/dataflag/orm.py#L251
```
df.DataFlag.select(df.DataFlagVote.lsd).distinct().join(df.DataFlagVote).where(df.DataFlagVote.revision == revision, df.DataFlag.type == "vote")
```
The result is empty (there are no DataFlagVote's in the db)

ch_pipeline/processing/quarterstack.py
jrs65: I think this is about the most minimal query you can do, the `.tuples()` call means it's returning a tuple of all the fields specified within the `.select(...)` for each row. As I only specify one field I need to pull it out of each 1-tuple. I don't think there's a cleaner way to do this.

ch_pipeline/processing/quarterstack.py
jrs65: From the `tag_pattern` spec in `DailyProcessing` I think it can only return strings which match the `\d+` regex, so they're guaranteed to be able to parse to integers. However, I do think it would be good to just parse `d` once so I can change that.

PR #74, opened 2021-03-05
feat(daily): new revision

jrs65:


ch_pipeline/processing/daily.py
```diff
@@ -242,19 +267,20 @@
         - "{catalogs[0]}"
         - "{catalogs[1]}"
         - "{catalogs[2]}"
+        - "{catalogs[3]}"
```
wulfda02, line 270: The CGRaBS catalog is entirely contained within the OVRO catalog, so there's no reason to keep CGRaBS around anymore

wulfda02 (dismissed):

nritsche (commented):
It seems there's a lot of reasoning behind each of these changes. Should we document these?

jrs65:
> It seems there's a lot of reasoning behind each of these changes. Should we document these?

Yes, I think that would be a good thing to do. In theory we keep a description of the revision in the root of each processed directory, but it's not super visible. Even a wiki page describing the changes between revisions would be helpful, this links in to the discussion we were having with @mondana about having better visibility and tracking of our pipeline products.

ssiegelx (dismissed):
This looks good.  The only comment I have is that we may want to add a `ThresholdVisWeight` after `SiderealRegridder` to reduce the re-gridding artifacts in the daily ringmaps and source spectra.

jrs65:
@nritsche do these commit messages gain your seal of approval?

nritsche:
> @nritsche do these commit messages gain your seal of approval?

Thanks! That's what I was hoping for.

jrs65:
We need a way of distinguishing them as they're not code changes. I think a config prefix is correct here, although something else (but not feat) might work. Is there any way of adjusting what auto-changelog does?

jrs65:
I think this is done. I've just changed all the config's to feat's in the commit messages. It's a bit of a hack, in the longer term we should figure out how to classify config changes like this, and make sure they hit the changelogs.

PR #76, opened 2021-04-26
refactor(core.telescope.CHIMEExternalBeam): Adapt to use new beam fil…

tristpinsm:
…e format.

Also makes frequency interpolation an option.

### DIFF
--- a/ch_pipeline/core/telescope.py
+++ b/ch_pipeline/core/telescope.py
@@ -17,6 +17,8 @@ from caput import config, mpiutil
 from drift.core import telescope
 from drift.telescope import cylbeam
 
+from draco.core.containers import ContainerBase
+
 from ch_util import ephemeris, tools
 
 
@@ -704,19 +706,31 @@ class CHIMEExternalBeam(CHIME):
     This class uses an external beam model that is read in from a file.
     """
 
-    primary_beamx_filename = config.Property(
-        proptype=str,
-        default="/project/k/krs/cahofer/pass1/beams/beamx_400_800_nfreq200.hdf5",
-    )
+    primary_beam_filename = config.Property(proptype=str)
+    freq_interp_beam = config.Property(proptype=bool, default=False)
 
-    primary_beamy_filename = config.Property(
-        proptype=str,
-        default="/project/k/krs/cahofer/pass1/beams/beamy_400_800_nfreq200.hdf5",
-    )
+    def _finalise_config(self):
+        """Get the beam file object."""
+        logger.debug("Reading beam model from {}...".format(self.primary_beam_filename))
+        self._primary_beam = ContainerBase.from_file(
+            self.primary_beam_filename, mode="r", distributed=False, ondisk=True
+        )
+
+        # cache axes
+        self._beam_freq = self._primary_beam.freq[:]
+        self._beam_nside = self._primary_beam.nside
+        # TODO must use bytestring here because conversion doesn't work with ondisk=True
+        self._beam_pol_map = {
+            "X": list(self._primary_beam.pol[:]).index(b"X"),
+            "Y": list(self._primary_beam.pol[:]).index(b"Y"),
+        }
+
+        if len(self._primary_beam.input) > 1:
+            raise ValueError("Per-feed beam model not supported for now.")
+
+        super()._finalise_config()
 
     def beam(self, feed, freq_id):
-        # Fetch beam parameters out of config database.
-
         feed_obj = self.feeds[feed]
         tel_freq = self.frequencies
         nside = self._nside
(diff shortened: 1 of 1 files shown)


tristpinsm:
black is failing on files that were not changed in this PR. Should I fix those here?

jrs65:
Yeah, I'm seeing that on other repos too. I think `black` 21.4 has been released.

jrs65:
So yes, please add a new `style` commit to fix them all across the repo.

ch_pipeline/core/telescope.py
```diff
@@ -532,19 +534,25 @@ class CHIMEExternalBeam(CHIME):
     This class uses an external beam model that is read in from a file.
     """
 
-    primary_beamx_filename = config.Property(
+    primary_beam_filename = config.Property(
         proptype=str,
         default="/project/k/krs/cahofer/pass1/beams/beamx_400_800_nfreq200.hdf5",
```
cahofer, line 539: this file doesn't exist anymore, we should link to the latest SVD primary beam model for example

ch_pipeline/processing/beam.py
```diff
@@ -121,7 +121,7 @@
 
 
 class HolographyFringestop(base.ProcessingType):
-    """"""
+    """ """
```
cahofer, line 124: why is this doctoring empty?

ch_pipeline/processing/daily.py
```diff
@@ -334,7 +334,7 @@
 
 
 class DailyProcessing(base.ProcessingType):
-    """"""
+    """ """
```
cahofer, line 338: same here ?

cahofer (commented):
Looks good to me except for the fact that we are pointing to a file that doesn't exist anymore.

ch_pipeline/core/telescope.py
tristpinsm: I guess it doesn't need a default value at all.

ch_pipeline/processing/beam.py
tristpinsm: these are not part of this PR. It's just bringing everything in line with the latest version of `black`.

ch_pipeline/core/telescope.py
```diff
@@ -554,27 +559,44 @@ def beam(self, feed, freq_id):
             raise ValueError("The requested feed doesn't seem to exist.")
 
         if tools.is_array_x(feed_obj):
-            fname = self.primary_beamx_filename
-
+            pol_ind = list(self._primary_beam.pol[:]).index(b"x")
```
jrs65, line 562: @tristpinsm I'm very surprised that you need to use byte strings here as I think that the dataset should be automatically converted to Unicode strings when loaded. 

Also, are these polarisation labels fixed, we tend to use upper case X and Y for polarisations, and I think that it's just confusing if we start to mix.

jrs65 (changes requested):

ch_pipeline/core/telescope.py
tristpinsm: the unicode conversion doesn't apply when you load the container with `ondisk=True`.

jrs65: Ah I see. Good point. Hmmm....

ch_pipeline/core/telescope.py
```diff
@@ -554,27 +559,44 @@ def beam(self, feed, freq_id):
...
         elif tools.is_array_y(feed_obj):
-            fname = self.primary_beamy_filename
-
+            pol_ind = list(self._primary_beam.pol[:]).index(b"y")
```
jrs65, line 564: @tristpinsm can you copy out the various ancillary datasets (i.e. freq and pol) into memory during the `_finalise_config`? Depending on how well HDF5 caches the datasets (and I think not well at all unless the dataset is chunked) this may end up re-reading them every time it needs to use them. For pol, you can even precompute the pol_ind's in there.

jrs65 (changes requested):
Thanks Tristan, this looks good, maybe just make the one change requested. I think this will work very well with the beam caching code that I've put into driftscan, so it shouldn't need to do much repeated IO even when processing large numbers of frequencies.

ch_pipeline/core/telescope.py
tristpinsm: ok, sounds good. I'll leave the bytestring in there for now, but I'll make and issue in `caput` and link to this code so we don't forget.

tristpinsm:
I'm going to quickly try this out on `cedar` now

tristpinsm:
> I'm going to quickly try this out on `cedar` now

done. Seems those last changes didn't break anything.

jrs65 (approved):

PR #78, opened 2021-05-10
feat(telescope): update beam evaluation for use in beamformer code

ssiegelx:
- Add angpos as a keyword argument to the beam method of the CHIME telescope.
Enables evaluation of the beam along a single source track, which is required
for the beamformer.  If not provided, then it will default to the _angos
telescope attribute.

- CHIME telescope now uses fast methods for evaluating the cylinder beam that
involve cacheing pre-products in a class dictionary.  This will require the
[sf/fast-beam](https://github.com/radiocosmology/driftscan/tree/sf/fast-beam) branch of driftscan to be merged.

- Add dec_normalized config parameter to CHIME class that, if set, will
normalize the beam by its magnitude at the transit of the declination
provided. To mach CHIME calibration, the user would set this parameter
to the declination of Cyg A.

- Creates a CHIMEParameterizedBeam class that is a CHIME telescope whose
beam method evaluates Mateus' parameterized fit to the driftscan beam.

### DIFF
--- a/ch_pipeline/analysis/calibration.py
+++ b/ch_pipeline/analysis/calibration.py
@@ -1811,7 +1811,7 @@ class ThermalCalibration(task.SingleTask):
         return gain
 
     def _ra2unix(self, csd, ra):
-        """ csd must be integer """
+        """csd must be integer"""
         return ephemeris.csd_to_unix(csd + ra / 360.0)
 
     def _reftime2gain(self, reftime_result, timestamp, frequency):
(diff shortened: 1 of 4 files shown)


sjforeman (dismissed):
Looks great to me!

jrs65:
Okay. This has all turned into a bit of a mess. Simon's changes don't use my optimisations, and thus this branch which shifts to use them for the CHIMETelescope, will bypass them kind of obviating the point.

I think I'll go and look at Simon's changes and see what's useful and try and get everything merged on the driftscan side. The hopefully we an get things merged here, as I have changes that need to be applied to ch-pipeline for

jrs65 (dismissed):

jrs65 (approved):

PR #82, opened 2021-05-26
feat(containers): set chunk sizes for zarr

nritsche:


ch_pipeline/processing/daily.py
```diff
@@ -201,7 +201,10 @@
       params:
         samples: 4096
         save: true
-        output_name: "sstream_{{tag}}.h5"
+        output_name: "sstream_{{tag}}.zarr"
+        output_format: zarr
```
anjakefala, line 205: The `output_name` suffix should be sufficient to describe the format.

ch_pipeline/processing/daily.py
```diff
@@ -201,7 +201,10 @@
...
+        output_compression: "bitshuffle"
+        output_compression_opts: [0, "lz4"]
```
anjakefala, line 207: We should set default compression options.

ch_pipeline/processing/daily.py
```diff
@@ -241,8 +242,29 @@
...
+      in: ringmap
+      out: ringmap_truncated
+      params:
         save: true
-        output_name: "ringmap_intercyl_{{tag}}.h5"
+        output_name: "ringmap_{{tag}}.zarr"
+        output_format: zarr
+        output_compression: "bitshuffle"
+        output_compression_opts: [0, "lz4"]
+    - type: draco.core.io.Truncate
```
anjakefala, line 257: Tristan made changes to `Truncate`, so we should check to make sure this config still holds.

ljgray:
@jrs65 I'm assuming this branch is outdated with the  changes you've since made to zarr, so I'd like to delete it if so

PR #83, opened 2021-06-04
[WIP] feat(core.telescope.CHIMEExternalBeam): Interpolate GridBeam.

tristpinsm:
This is not tested yet, but everything that is needed to do the interpolation is in. I thought I'd solicit some feedback early so we can figure out if this is how we want to move forward.

I implemented the interpolation in the `CHIMEExternalBeam` class because it sounded like that is where it would be most useful right now, but I think it would be straightforward to put it in a standalone class that could also be used by a task to produce `HEALPixBeam` containers (not sure where it would be appropriate to put such a class).

And would it be useful to cache the beam maps as they are produced, or do they only get used once?

### DIFF
--- a/ch_pipeline/core/telescope.py
+++ b/ch_pipeline/core/telescope.py
@@ -11,13 +11,16 @@ import logging
 import numpy as np
 import h5py
 import healpy
+from scipy.interpolate import RectBivariateSpline
 
 from caput import config, mpiutil
+
+from cora.util import coord, hputil
 
 from drift.core import telescope
 from drift.telescope import cylbeam
 
-from draco.core.containers import ContainerBase
+from draco.core.containers import ContainerBase, GridBeam, HEALPixBeam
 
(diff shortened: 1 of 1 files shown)


tristpinsm:
OK, I tested this out and with some changes it appears to be working.

tristpinsm:
Oh, and I should clarify that I'm not advocating we settle on this solution in the long run, as it's a bit of a hack as I mentioned. I wanted to put the code together in one place so we can see what it takes to do the interpolation, and then consider what would be best in the future.

If we decide we want to do the interpolation on the fly in the telescope object, then I'd suggest we decide on a standard for the `GridBeam` files that is closer to the `HEALPixBeam`, i.e. with the polarisation projection patterns and other massaging done offline. That would also simplify the interpolation code and make it easier to move it somewhere it could be reused.

jrs65 (dismissed):

jrs65 (approved):

PR #84, opened 2021-06-04
Frequency dependent FWHMs

saurabh-astro:
Based on FWHM computed from point source model (585-800 MHz), this PR revises north-south FWHM being used in the driftscan beam. Beam measurements point to a lower beam width compared to the existing beam width in driftscan. The discrepancy is larger in X pol. In order to avoid ripples in the beam, FWHM (as a function of frequency) computed for the point source model is fit with a cubic polynomial, and coefficients of the fit are provided to `fwhm_e` and `fwhm_h` functions.

Note that FWHM was optimized only in 585-800 MHz. Using it below that will result in errors due to extrapolation. 

Here is the N-S beam and FWHM comparisons between the two models:

FWHM_power
all_models_post_power

Per frequency comparisons are here: https://bao.chimenet.ca/doc/documents/1448

### DIFF
--- a/ch_pipeline/core/telescope.py
+++ b/ch_pipeline/core/telescope.py
@@ -197,7 +197,9 @@ class CHIME(telescope.PolarisedTelescope):
     def fwhm_e(self):
         """Full width half max of the E-plane antenna beam."""
 
-        coeff_fwhm_input_Y = np.array([1.15310483e-07,-2.30462590e-04,1.50451290e-01,-3.07440520e+01])
+        coeff_fwhm_input_Y = np.array(
+            [1.15310483e-07, -2.30462590e-04, 1.50451290e-01, -3.07440520e01]
+        )
 
         fwhm_freq_Y = np.polyval(coeff_fwhm_input_Y, self.frequencies)
 
@@ -207,7 +209,9 @@ class CHIME(telescope.PolarisedTelescope):
     def fwhm_h(self):
         """Full width half max of the H-plane antenna beam."""
 
-        coeff_fwhm_input_X = np.array([2.97495306e-07,-6.00582101e-04,3.99949759e-01,-8.66733249e+01])
+        coeff_fwhm_input_X = np.array(
+            [2.97495306e-07, -6.00582101e-04, 3.99949759e-01, -8.66733249e01]
+        )
 
         fwhm_freq_X = np.polyval(coeff_fwhm_input_X, self.frequencies)
 
--- a/ch_pipeline/core/telescope.py
+++ b/ch_pipeline/core/telescope.py
@@ -196,9 +196,9 @@ class CHIME(telescope.PolarisedTelescope):
     @property
     def fwhm_e(self):
         """Full width half max of the E-plane antenna beam."""
-        coeff_fwhm_input_Y = np.array([2.41511275e-10,-6.71378830e-07,\
-        7.20183967e-04,-3.52472912e-01,6.73758644e+01])
-        
+
+        coeff_fwhm_input_Y = np.array([1.15310483e-07,-2.30462590e-04,1.50451290e-01,-3.07440520e+01])
+
         fwhm_freq_Y = np.polyval(coeff_fwhm_input_Y, self.frequencies)
 
         return fwhm_freq_Y
@@ -206,9 +206,9 @@ class CHIME(telescope.PolarisedTelescope):
     @property
     def fwhm_h(self):
         """Full width half max of the H-plane antenna beam."""
-        coeff_fwhm_input_X = np.array([ 2.86443038e-10,-7.79063706e-07,\
-        8.15859136e-04,-3.89137634e-01,7.23010129e+01])
-        
+
+        coeff_fwhm_input_X = np.array([2.97495306e-07,-6.00582101e-04,3.99949759e-01,-8.66733249e+01])
+
         fwhm_freq_X = np.polyval(coeff_fwhm_input_X, self.frequencies)
 
         return fwhm_freq_X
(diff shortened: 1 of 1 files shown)


ssiegelx:
Hey Saurabh, this looks great!

My only question is how does the polynomial behave below 585 MHz?  I suspect that it will blow up if it's not being constrained at those frequencies.  I think we need to make sure that we still have the ability to simulate a somewhat reasonable beam model in the lower portion of the band.

One option would be to move the polynomial coefficient definition to class attributes defined right below the config properties.  For example

```
    # Fix base properties
    cylinder_width = 20.0
    cylinder_spacing = tools._PF_SPACE

    _ewidth = [0.7]
    _hwidth = [1.2]
```

Then reference these class attributes in the `fwhm_e` and `fwhm_h` properties:

```
@property
def fwhm_h(self):
    return np.polyval(np.array(self._ewidth) * 2.0 * np.pi / 3.0,  self.frequencies)
```

Then create a new subclass of `CHIME` that only overwrites the `_ewidth` and `_hwidth` attributes.  At least this way people can continue using the old version if they need to simulate the lower part of the band, or the new class if they are only focusing on the upper part.

Another option would be to somehow constrain the polynomial at frequencies below 585 MHz.  Long term I can generate the point source beam at these frequencies, but until then perhaps we could use the SVD beam?  Is the N-S FHWM of the SVD beam comparable to the point source beam in the upper part of the band?

saurabh-astro:
> Hey Saurabh, this looks great!
> 
> My only question is how does the polynomial behave below 585 MHz? I suspect that it will blow up if it's not being constrained at those frequencies. I think we need to make sure that we still have the ability to simulate a somewhat reasonable beam model in the lower portion of the band.
> 
> One option would be to move the polynomial coefficient definition to class attributes defined right below the config properties. For example
> 
> ```
>     # Fix base properties
>     cylinder_width = 20.0
>     cylinder_spacing = tools._PF_SPACE
> 
>     _ewidth = [0.7]
>     _hwidth = [1.2]
> ```
> 
> Then reference these class attributes in the `fwhm_e` and `fwhm_h` properties:
> 
> ```
> @property
> def fwhm_e(self):
>     return np.polyval(np.array(self._ewidth) * 2.0 * np.pi / 3.0,  self.frequencies)
> ```
> 
> Then create a new subclass of `CHIME` that only overwrites the `_ewidth` and `_hwidth` attributes with the best fit coefficients. At least this way people can continue using the old telescope class if they need to simulate the lower part of the band, or the new class if they are only focusing on the upper part.
> 
> Another option would be to somehow constrain the polynomial at frequencies below 585 MHz. Long term I can generate the point source beam at these frequencies, but until then perhaps we could use the SVD beam? Is the N-S FHWM of the SVD beam comparable to the point source beam in the upper part of the band?

Thanks, that makes sense. Extrapolation does not produce reasonable results below 500 MHz. I will have a look at SVD response. Some comparisons in the past showed it to match the mean profile, so we could potentially use that. If there are issues with it, I will modify the code to preserve the default parameters.

jrs65:
@saurabh-astro I think the way this is implemented it will effect the beam widths in the x direction as well (although it'll go through a diffraction calculation). Do we think that these changes are reasonable in the x direction? I appreciate we don't know that much about it, but it would be good to think about if that makes sense.

ch_pipeline/core/telescope.py
```diff
@@ -196,12 +199,14 @@ def _finalise_config(self):
     @property
     def fwhm_e(self):
```
jrs65, line 200: You should probably use `caput.cache.cahed_property` as the decorator here to avoid redoing the polynomial fit every time it is referenced.

ch_pipeline/core/telescope.py
```diff
@@ -526,6 +531,32 @@ def _unique_beams(self):
         return beam_map, beam_mask
 
 
+class revisedDriftScan(CHIME):
```
jrs65, line 534: Probably think of a better name here.

jrs65 (changes requested):

saurabh-astro:
It would certainly change x profile. I will run the default and revised models to assess the difference.

ch_pipeline/core/telescope.py
saurabh-astro: Thanks, I will have a look at it.

saurabh-astro:
The last set of FWHMs changed the EW profiles such that the response was mismatched when compared with holography sources. As a quick fix, the current version uses different E and H plane widths for the two polarizations. This ensures that the beam FWHM is consistent with NS profile from point source model as well as EW response from holography. The relevant plots are here: https://bao.chimenet.ca/doc/documents/1448/3/get/ds_revised.pdf?n=4971

ch_pipeline/core/telescope.py
```diff
@@ -194,14 +200,28 @@ def _finalise_config(self):
 
     # Tweak the following two properties to change the beam width
     @property
-    def fwhm_e(self):
+    def fwhm_ex(self):
         """Full width half max of the E-plane antenna beam."""
```
ssiegelx, line 204:
```suggestion
        """Full width half max of the E-plane antenna beam for the X polarization."""
```
and likewise for the other 3 properties.

ssiegelx (approved):

jrs65:
@saurabh-astro this is looking good to me. I think you just need to implement Seth's suggestions and my small ones from the last round (`cached_property` and renaming the class). I'm not sure what a good name for the new class is, maybe `CHIMERevisedBeam`?

saurabh-astro:
Okay, I have made the some documentation edits and a few minor changes.

jrs65 (approved):

jrs65:
@saurabh-astro looks good to me. Please squash merge it down with an appropriately formatted commit message.

PR #85, opened 2021-06-23
Frequency dependent FWHMs (#84)

saurabh-astro:
feat(telescope): adds a new subclass of `CHIME`: `CHIMEFitBeam`.

It revises E and H plane FWHMs for X and Y pols separately for the dipole beam.
The numbers are  optimized using point source beam model in N-S direction and
holography in E-W direction.


saurabh-astro:
Sorry, this was not supposed to be a separate PR. Closing this one.

PR #86, opened 2021-07-12
feat(source_removal): reduce memory usage and other improvements

ssiegelx:
- Use new ephemeris methods for computing source positions.

- Add capability to restrict the range of hour angles covered by the source model through the min_ha and max_ha config properties.

- Allow separate masking of N-S and E-W components of the baseline.

- Change data type of the param index map of SourceModel and remove the redundant source_index dataset.

- Move calculation of the source model into the frequency and polarisation loops to reduce memory usage.

### DIFF
--- a/ch_pipeline/analysis/source_removal.py
+++ b/ch_pipeline/analysis/source_removal.py
@@ -100,6 +100,8 @@ def model_extended_sources(
     ntime = timestamp.size
     nbaseline = distance.shape[-1]
 
+    ones = np.ones((nfreq, nbaseline, ntime), dtype=np.float64)
+
     # Calculate baseline distances in wavelengths
     lmbda = speed_of_light * 1e-6 / freq
     u, v = (
@@ -148,9 +150,9 @@ def model_extended_sources(
         # Evaluate polynomial
         aa, bb = source_bound[ss], source_bound[ss + 1]
 
-        H = np.polynomial.hermite.hermvander3d(
-            u * scale[ss, 0], v * scale[ss, 1], ha * scale[ss, 2], poly_deg[ss]
-        )
+        coords = [ax * scale[ss, ii] * ones for ii, ax in enumerate([u, v, ha])]
+
+        H = np.polynomial.hermite.hermvander3d(*coords, poly_deg[ss])
 
         # Calculate the fringestop phase
         phi = tools.fringestop_phase(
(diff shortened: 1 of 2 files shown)


ch_pipeline/analysis/source_removal.py
```diff
@@ -920,13 +1080,149 @@ def process(self, data, model, beams):
...
+
+        coef = np.sum(t, axis=0)
+
+    assert coef.size == k * (m - 1) + 1
+
+    return coef / m ** k
+
+
+def apply_kz_lpf_2d(y, flag, window=3, niter=8, mode="wrap", frac_required=0.80):
+    """Apply a 2D Kolmogorov-Zurbenko filter.
```
jrs65, line 1147: I appreciate I'm being a bit lazy, but it might be good to have a sentence about what a Kolmogorov-Zurbenko filter actually does for the ignorant (like myself).

sjforeman: If you visit the Wikipedia page, you will be rewarded with some nice photos of Kolmogorov and Zurbenko themselves: https://en.wikipedia.org/wiki/Kolmogorov–Zurbenko_filter

jrs65: 🤣 

That's hilarious. It's like their holiday photos or something, particularly the of them on the ship looking at handbags!

ch_pipeline/analysis/source_removal.py
```diff
@@ -484,6 +538,89 @@ def process(self, data):
...
+        Parameters
+        ----------
+        model : containers.SourceModel
+            Best-fit parameters of the source model.
+
+        Returns
+        -------
+        model : containers.SourceModel
+            The input container with the amplitude dataset
+            filtered and the weights updated.
```
saurabh-astro, line 588: May be I missed it, but where are weights being updated in this function?

saurabh-astro (dismissed):
Looks great! I just had a minor question, otherwise this looks good to go.

ch_pipeline/analysis/source_removal.py
ssiegelx: I've updated the doctoring to give a better explanation.

ssiegelx:
I've addressed the comments and will go ahead and merge this now.

PR #87, opened 2021-07-12
feat(inject): tasks for injecting catalogs into visibility data

ssiegelx:
Introduces the FluxCatalog and SpectroscopicCatalog tasks,
which generate synthetic visibilities from a catalog of sources.
Also introduces versions of these tasks that modulate the signal
from each source by a beam model, either an external model, or
one implemented via the beam method of the provided telescope object.

Introduces code for perturbing the redshifts of the eBOSS tracers.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,6 @@ jobs:
 
     - name: Install pip dependencies
       run: |
-        pip install --use-deprecated=legacy-resolver git+https://github.com/chime-experiment/ch_util-stub.git
         pip install future
         pip install numpy
         pip install cython
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,12 +44,12 @@ jobs:
 
     - name: Install pip dependencies
       run: |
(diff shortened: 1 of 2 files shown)


ssiegelx:
This is a stand-alone module that I have been using to synthesize visibilities for catalogs of point sources.  This capability is used in a few places:

- Solving for the beam model by fitting hybrid visibilities from intercylinder baselines to point source models of the sky
- Predicting the signal to noise of the eBOSS stacks by injecting into CHIME data
- Characterizing the transfer function of the stacking pipeline
    
I've placed it in `ch_pipeline` rather than `draco` because it requires the `ch_util.fluxcat` module.

It also contains functions for perturbing the redshift of sources using the velocity error distributions for the eBOSS tracers (described here: https://github.com/chime-experiment/Stacking/issues/14). We may want to move those functions to a module in `draco.util`, either tools or something new, since we will need to perturb the redshifts elsewhere.

ssiegelx:
Merging this older PR. It’s self-contained, broadly useful, and has already been used extensively.

PR #91, opened 2021-08-26
feat(timing): ApplyTimingCorrection only to data flagged by needs_timing_correction

### LINKED ISSUE #80: Fix timing correction to work with most recent data
At the moment the `ApplyTimingCorrection` task fails when running on more recent data that doesn't need it at all. We need to fix it such that it simply doesn't apply anything on this data.

Seth suggested that the best way to do this would be to use a dataflag to indicate the range of time over which a timing correction should in theory be applied, although it might be easier to just hard code the switch over data in the code.

anjakefala:
This is ready for review. In addition to the code changes, I want to confirm that the date range I proposed makes sense, and discuss if we want to modify the task `DataFlagger` so its default does not include flags like `needs_timing_correction`. 

- [x] Testdaily on days that need timing correction
- [x] Testdaily on days that do not need timing correction
- [x] Potentially adjust `DataFlagger` task to never mask based on `needs_timing_correction` flag

```
id: None
type: needs_timing_correction
start: 2018-10-11 14:00:00+00:00
finish: 2020-03-04 01:13:01+00:00
metadata: 
    {instrument: chime, user: Anja Boskovic}
```

Closes #80

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -1703,10 +1703,22 @@ class DataFlagger(task.SingleTask):
     Parameters
     ----------
     flag_type : list
-        List of DataFlagType names to apply. Defaults to 'all'
+        List of DataFlagType names to apply. Defaults to the flags representing ranges of time known to be bad
+        that may effect the delay spectrum estimate.
     """
 
-    flag_type = config.Property(proptype=list, default=["all"])
+    flag_type = config.list_type(
+        type_=str,
+        default=[
+            "acjump",
+            "bad_calibration_acquisition_restart",
+            "bad_calibration_fpga_restart",
+            "bad_calibration_gains",
+            "decorrelated_cylinder",
+            "globalflag",
+            "rain1mm",
+        ],
+    )
 
     def setup(self):
         """Query the database for flags of the requested types."""
--- a/ch_pipeline/analysis/timing.py
+++ b/ch_pipeline/analysis/timing.py
@@ -6,12 +6,19 @@ import os
 import numpy as np
 
 from ch_util import timing, ephemeris
-from caput import config, pipeline
+from caput import config
+from caput.pipeline import PipelineRuntimeError
 from draco.core import task
+
+# For querying DataFlag database
+from chimedb import dataflag as df
+from chimedb.core import connect as connect_database
 
 
 class ApplyTimingCorrection(task.SingleTask):
     """Apply a timing correction to the visibilities.
+
+    Only runs on days flagged by needs_timing_correction.
 
     Parameters
     ----------
@@ -45,6 +52,37 @@ class ApplyTimingCorrection(task.SingleTask):
             Note that the timing correction must *already* be referenced with respect
             to the times that were used to calibrate if processing stacked data.
         """
+        flags = None
+
+        # Query flag database for date ranges that need timing correction
+        if self.rank == 0:
+            connect_database()
+
+            needs_timing_correction_type = df.DataFlagType.get(
+                name="needs_timing_correction"
+            )
+
+            if needs_timing_correction_type is None:
+                raise RuntimeError(
+                    "Could not find data flag type `needs_timing_correction`"
+                )
+
+            # We are extracting the time ranges here from the DataFlag objects
+            # to avoid potential issues with non-0 ranks accidentally doing database queries later on
+            flags = list(
+                df.DataFlag.select(df.DataFlag.start_time, df.DataFlag.end_time)
+                .where(df.DataFlag.type == needs_timing_correction_type)
+                .dicts()
+            )
+
+            self.log.debug(
+                f"Queried database and received {len(flags)} needs_timing_correction flags."
+            )
+
+        # Share timing correction flags with other nodes
+        # Save flags to class attribute
+        self.flags = self.comm.bcast(flags, root=0)
+
         if not isinstance(tcorr, list):
             tcorr = [tcorr]
         self.tcorr = tcorr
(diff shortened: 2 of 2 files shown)


ch_pipeline/analysis/timing.py
```diff
@@ -45,6 +51,36 @@ def setup(self, tcorr):
...
+            if needs_timing_correction_type is None:
+                raise RuntimeError(
+                    "Could not find data flag type `needs_timing_correction`"
+                )
+
+            needs_timing_correction_flags = list(
+                df.DataFlag.select().where(
+                    df.DataFlag.type == needs_timing_correction_type
+                )
+            )
```
jrs65, line 73: You may want to extract the time ranges here and use those instead of the DataFlag objects. The potential issue is that if you have a DataFlag object and do `df.type` it may go off and try and do a database query to get the type object. That's fine on rank=0 as you have a DB connection, but not on the other ranks. We may make that mistake elsewhere in the pipeline.

ch_pipeline/analysis/timing.py
```diff
@@ -45,6 +51,36 @@ def setup(self, tcorr):
...
+
+            self.log.info(
+                "Queried database and received %d needs_timing_correction flags."
+                % len(needs_timing_correction_flags)
```
jrs65, line 77: Probably f-string this one.

ch_pipeline/analysis/timing.py
```diff
@@ -83,6 +119,33 @@ def process(self, tstream):
...
+                    % (
+                        timestamp[0],
+                        timestamp[-1],
+                        ephemeris.unix_to_datetime(flag.start_time).strftime(
+                            "%Y%m%dT%H%M%SZ"
+                        ),
+                        ephemeris.unix_to_datetime(flag.finish_time).strftime(
+                            "%Y%m%dT%H%M%SZ"
+                        ),
+                    )
```
jrs65, line 138: Maybe another f-string here.

ch_pipeline/analysis/timing.py
```diff
@@ -83,6 +119,33 @@ def process(self, tstream):
...
+                )
+                needs_timing_correction = True
+
+        if not needs_timing_correction:
+            self.log.info(
+                "Sample in span %d to %d does not need timing correction"
```
jrs65, line 144: Also here.

ch_pipeline/analysis/timing.py
```diff
@@ -83,6 +119,33 @@ def process(self, tstream):
         # If requested, extract the input flags
         input_flags = tstream.input_flags[:] if self.use_input_flags else None
 
+        # Check needs_timing_correction flags time ranges to see if input timestream
+        # falls within the interval
+        needs_timing_correction = False
+        for flag in self.flags:
+            if timestamp[0] >= flag.start_time and timestamp[0] <= flag.finish_time:
```
jrs65, line 126: You may want to check for the case that the timing correction needed status may change within a file. Not sure what to do in that case (probably reject it), but whatever you do a clear logging statement would be good.

ch_pipeline/analysis/timing.py
```diff
@@ -45,6 +51,36 @@ def setup(self, tcorr):
             Note that the timing correction must *already* be referenced with respect
             to the times that were used to calibrate if processing stacked data.
         """
+        # Query flag database for date ranges that need timing correction
+        needs_timing_correction_flags = None
+
+        if mpiutil.rank0:
```
jrs65, line 57: You should use `self.rank == 0` in here. Generally I'd try to stay away from `mpiutil` as much as you can, it's designed to help writing code in contexts where you might not have MPI installed, but I've found it's more confusion than it's worth. Code in `draco` you can basically assume MPI will be available if you need it, and so that functionality is baked into `SingleTask`.

jrs65 (changes requested):
Looking good! Just a few comments.

ch_pipeline/analysis/timing.py
anjakefala: I am not sure what you mean by reject it? Does it mean not apply the timing correction? That is what I did for now!

anjakefala:
All requested adjustments made! Uncertainty for what to do about the border case. Currently, I do not apply the timing correction, and I got the issue.

ch_pipeline/analysis/timing.py
```diff
@@ -45,6 +52,48 @@ def setup(self, tcorr):
...
+
+            nflags = len(needs_timing_correction_flags)
+            if nflags == 0:
+                raise RuntimeError(
+                    "Database query should return at least one needs_timing_correction flag."
+                )
+
+            self.log.info(
+                f"Queried database and received {nflags} needs_timing_correction flags."
+            )
```
jrs65, line 84: Probably more like a debug message.

ch_pipeline/analysis/timing.py
```diff
@@ -45,6 +52,48 @@ def setup(self, tcorr):
...
+
+            # We are extracting the time ranges here from the DataFlag objects
+            # to avoid potential issues with non-0 ranks accidentally doing database queries later on
+            flags = [
+                {"start_time": flag.start_time, "finish_time": flag.finish_time}
+                for flag in needs_timing_correction_flags
+            ]
```
jrs65, line 91: I think you can probably remove this in favour of either using the DataFlag instances directly, or just using `.dict()` on the query to fetch dictionaries originally rather than DataFlag instances.

ch_pipeline/analysis/timing.py
```diff
@@ -83,6 +132,34 @@ def process(self, tstream):
...
+                        f"needs_timing_correction DataFlag covering {ephemeris.unix_to_datetime(flag['start_time']).strftime('%Y%m%dT%H%M%SZ')} "
+                        f"to {ephemeris.unix_to_datetime(flag['finish_time']).strftime('%Y%m%dT%H%M%SZ')}."
+                    )
+                else:
+                    self.log.info(
+                        f"Sample covering {timestamp[0]} to {timestamp[-1]} flagged by "
+                        f"needs_timing_correction DataFlag covering {ephemeris.unix_to_datetime(flag['start_time']).strftime('%Y%m%dT%H%M%SZ')} "
+                        f"to {ephemeris.unix_to_datetime(flag['finish_time']).strftime('%Y%m%dT%H%M%SZ')}. Timing correction will be applied."
+                    )
+                    needs_timing_correction = True
```
jrs65, line 155: This should probably `break` at this point.

ch_pipeline/analysis/timing.py
```diff
@@ -83,6 +132,34 @@ def process(self, tstream):
...
+
+        if not needs_timing_correction:
+            self.log.info(
+                f"Sample in span {timestamp[0]} to {timestamp[-1]} does not need timing correction"
```
jrs65, line 159: Probably call it data rather than sample.

Also, if you're giving formatted datetimes elsewhere you should probably give them here too.

jrs65 (changes requested):
Looks good, just a few minor requests.

ch_pipeline/analysis/timing.py
jrs65: Haha. I note that I suggested to not use the DataFlag instances directly. Ooops. In that case maybe just use `.dict()`(`.namedtuple()` would probably also do).

ch_pipeline/analysis/flagging.py
```diff
@@ -1703,10 +1703,22 @@ class DataFlagger(task.SingleTask):
     Parameters
     ----------
     flag_type : list
-        List of DataFlagType names to apply. Defaults to 'all'
+        List of DataFlagType names to apply. Defaults to the flags representing ranges of time known to be bad
+        that may effect the delay spectrum estimate.
     """
 
-    flag_type = config.Property(proptype=list, default=["all"])
+    flag_type = config.Property(
```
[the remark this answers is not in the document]
anjakefala: Makes sense! We do not need the additional power of `maxlength` or `length`, but being able to define the expected type of the content of the list is useful.

ch_pipeline/analysis/timing.py
anjakefala: > just using .dict() on the query to fetch dictionaries originally rather than DataFlag instances.

I do not understand what you mean by this. =( Because, it is potentially returning a list of instances, so I am not sure where the `dict()` would go. Do you mean

```
flags = [dict(flag) for flag in needs_timing_correction_flags]
    
```

jrs65: Ah, no, sorry, that wasn't clear, especially as I had a typo. I mean use this: http://docs.peewee-orm.com/en/latest/peewee/api.html#BaseQuery.dicts

So you could effectively just do:
```python
flags = list(df.DataFlag.select(df.DataFlag.start_time, df.DataFlag.end_time).where(
    df.DataFlag.type == needs_timing_correction_type
).dicts())
```
and you'd get exactly the same thing, but with peewee doing all the lifting for you, and skipping a bunch of intermediate object construction it no longer needs to do.

jrs65 (approved):

PR #93, opened 2021-12-10
Changes for rev_04

anjakefala:
Includes:
* Only start tasks for days which actually have data available online
* Adjust the daily pipeline to process available days up until October 2020
* Queue one day every 7, and then cycle back, to prioritise collecting a wider breadth

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -1,4 +1,10 @@
+import os
+import math
 import datetime
+
+import chimedb.core as db
+from ch_util import ephemeris
+import chimedb.data_index as di
 
 from . import base
 
@@ -403,20 +409,85 @@ class DailyProcessing(base.ProcessingType):
         for t in self._revparams["intervals"]:
             self._intervals.append((t["start"], t.get("end", None), t.get("step", 1)))
 
-
         self._padding = self._revparams["padding"]
+
+    def _available_files(self, start_csd, end_csd):
+        """
+        Return chimestack files available in cedar_online between start_csd and end_csd, if all of the files for that period are available online.
+
+        Return an empty list if files between start_csd and end_csd are only partially available online.
+
+        Total file count is verified by checking files that exist everywhere.
+
+        Parameters
+        ----------
+        start_csd : int
+            Start date in sidereal day format
+        end_csd : int
+            End date in sidereal day format
+
+        Returns
+        -------
+        list
+            List contains the chimestack files available in the timespan, if all of them are available online
+
+        """
+
+        # Connect to databases
+        db.connect()
+
+        # Get timestamps in unix format
+        # Needed for queries
+        start_time = ephemeris.csd_to_unix(start_csd)
+        end_time = ephemeris.csd_to_unix(end_csd)
+
+        # We will want to know which files are in chime_online and nearline on cedar
+        online_node = di.StorageNode.get(name="cedar_online", active=True)
+        chimestack_inst = di.ArchiveInst.get(name="chimestack")
+
+        # TODO if the time range is so small that it’s completely contained within a single file, nothing will be returned
+        # have to special-case it by looking for files which start before the start time and end after the end time).
+
+        archive_files = (
+            di.ArchiveFileCopy.select(
+                di.CorrFileInfo.start_time,
+                di.CorrFileInfo.finish_time,
+            )
+            .join(di.ArchiveFile)
+            .join(di.ArchiveAcq)
+            .switch(di.ArchiveFile)
+            .join(di.CorrFileInfo)
+        )
+
+        # chimestack files available online which include between start and end_time
+
+        files_that_exist = archive_files.where(
+            di.ArchiveAcq.inst
+            == chimestack_inst,  # specifically looking for chimestack files
+            di.CorrFileInfo.start_time
+            < end_time,  # which contain data that includes start time and end time
+            di.CorrFileInfo.finish_time >= start_time,
+            di.ArchiveFileCopy.has_file == "Y",
+        )
+
+        files_online = files_that_exist.where(
+            di.ArchiveFileCopy.node == online_node,  # that are online
+        )
+
+        filenames_online = sorted([t for t in files_online.tuples()])
+
+        # files_that_exist might contain the same file multiple files
+        # if it exists in multiple locations (nearline, online, gossec, etc)
+        # we only want to include it once
+        filenames_that_exist = sorted(list(set(t for t in files_that_exist.tuples())))
+
+        return filenames_online, filenames_that_exist
 
     def _available_tags(self):
         """Return all the tags that are available to run.
 
         This includes any that currently exist or are in the job queue.
         """
-
-        # TODO: should decide availability based on what data is actually
-        # available
-        # - Need to find all correlator data in the range
-        # - Figure out which ones are on cedar
-        # - Figure out which sidereal days are covered by this range
 
         csds = []
 
@@ -426,9 +497,88 @@ class DailyProcessing(base.ProcessingType):
             csd_set = set(csds)
             csds += [csd for csd in csd_i if csd not in csd_set]
 
+        # grab the list of files that are online, and that exist anywhere, from the earliest csd to the latest
+        csds_sorted = sorted(csds)
+        filenames_online, filenames_that_exist = self._available_files(
+            csds_sorted[0], csds_sorted[-1] + 1
+        )
+
+        # only queue jobs for which all data is available online in chime_online
+        csds_available = self._csds_available_data(
+            csds_sorted, filenames_online, filenames_that_exist
+        )
+        csds = [csd for csd in csds if csd in csds_available]
+
         tags = ["%i" % csd for csd in csds]
 
         return tags
+
+    def _csds_available_data(self, csds, filenames_online, filenames_that_exist):
+        """
+        Return the subset of csds in `csds` for whom all files are online.
+
+        `filenames_online` and `filenames_that_exist` are a list of tuples
+        (start_time, finish_time)
+
+        All 3 lists should be sorted.
+        """
+        csds_available = []
+
+        for csd in csds:
+            start_time = ephemeris.csd_to_unix(csd)
+            end_time = ephemeris.csd_to_unix(csd + 1)
+
+            # online - list of filenames that are online between start_time and end_time
+            # index_online, the final index in which data was located
+            online, index_online = self._files_in_timespan(
+                start_time, end_time, filenames_online
+            )
+            exists, index_exists = self._files_in_timespan(
+                start_time, end_time, filenames_that_exist
+            )
+
+            if (len(online) == len(exists)) and (len(online) != 0):
+                csds_available.append(csd)
+
+            # The final file in the span may contain more than one sidereal day
+            index_online = max(index_online - 1, 0)
+            index_exists = max(index_exists - 1, 0)
+
+            filenames_online = filenames_online[index_online:]
+            filenames_that_exist = filenames_that_exist[index_exists:]
+
+        return csds_available
+
+    def _files_in_timespan(self, start, end, files):
+        """
+        Parameters
+        ----------
+        start : float
+            unix timestamp
+        end : float
+            unix timestamp
+        files : list of tuple
+            tuple: (start_time, finish_time)
+
+        Returns
+        -------
+        list of elements of `files`, whose start_time and finish_time
+        fall between `start` and `end`
+
+        index of the first file whose `start_time` is after `end`
+        """
+        available = []
+        for i in range(0, len(files)):
+            f = files[i]
+            if (f[0] < end) and (f[1] >= start):
+                available.append(f)
+            # files are in chronological order
+            # once we hit this conditional, there are no files
+            # further in the list, which will fall within
+            # our timewindow
+            elif f[0] > end:
+                return available, i
+        return available, len(files) - 1
 
     def _finalise_jobparams(self, tag, jobparams):
         """Set bounds for this CSD."""
@@ -485,9 +635,6 @@ def csds_in_range(start, end, step=1):
     -------
     csds : list of ints
     """
-
-    import math
-    from ch_util import ephemeris
 
     if end is None:
         end = datetime.datetime.utcnow()
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -353,20 +353,15 @@ class DailyProcessing(base.ProcessingType):
             # A good looking interval from late 2019 (determined from run
             # notes, dataflags and data availability)
             {"start": "CSD2143", "end": "CSD2148"},
-            ## Runs processed for rev_00
-            # October run
-            {"start": "20181011T140000Z", "end": "20181019T220000Z"},
-            # Winter run periods - as defined by Mateus
-            # Pass A (start trimmed for timing solutions)
-            {"start": "20181223T000000Z", "end": "20181229T000000Z"},
-            # Pass B
-            {"start": "20190111T000000Z", "end": "20190207T000000Z"},
-            # Pass C (end trimmed for timing sols)
-            {"start": "20190210T000000Z", "end": "20190304T000000Z"},
-            # April run
-            {"start": "20190406T000000Z", "end": "20190418T000000Z"},
-            # July run
-            {"start": "20190713T000000Z", "end": "20190723T000000Z"},
+            # intervals for which we process only one day every 7 days
+            ## dataflags and calibration tables are currently only available until October 2020
+            {"start": "CSD1878", "end": "CSD2539", "step": 7},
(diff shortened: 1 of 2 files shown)


ch_pipeline/processing/daily.py
```diff
@@ -353,20 +359,17 @@ class DailyProcessing(base.ProcessingType):
...
+        # intervals for which we process only one day every 7 days
+        "intervals_weekly": [
+            ## dataflags and calibration tables are currently only available until October 2020
+            {"start": "20181011T000000Z", "end": "20201021T220000Z"},
+            {"start": "20181012T000000Z", "end": "20201021T220000Z"},
+            {"start": "20181013T000000Z", "end": "20201021T220000Z"},
+            {"start": "20181014T000000Z", "end": "20201021T220000Z"},
+            {"start": "20181015T000000Z", "end": "20201021T220000Z"},
+            {"start": "20181016T000000Z", "end": "20201021T220000Z"},
+            {"start": "20181017T000000Z", "end": "20201021T220000Z"},
```
jrs65, line 372: I think it would be cleaner here to allow these specs to have a `step` field, and just put these into the main intervals list with `"step": 7`, and remove the `intervals_weekly` entirely.

Also, these definitely should be defined in terms of their CSD and not their calendar date. Over the course of a year sidereal and calendar days shift by a whole year with respect to each other, so you will get off the expected schedule at some point.

ch_pipeline/processing/daily.py
```diff
@@ -407,29 +410,136 @@ def _load_hook(self):
...
+            == "chimestack",  # specifically looking for chimestack fiels
+            di.CorrFileInfo.start_time
+            < end_time,  # which contain data that includes start time and end time
+            di.CorrFileInfo.finish_time >= start_time,
+            di.ArchiveFileCopy.node == online_node,  # that are online
+            di.ArchiveFileCopy.has_file == "Y",
+        )
+
+        # chimestack files available on nearline, which include between start and end_time
+        files_nearline = archive_files.where(
```
jrs65, line 487: I think comparing to the contents of nearline isn't quite the right comparison, you need to compare to all available files.

Example of where this breaks: a day just being transferred over, all files are on gong, some are on online and some are on nearline. In this situation all the files in nearline are probably online, so your logic would start processing, but it shouldn't do anything until the files from gong are available.

The resolution is probably to compare to:
- Remove the node == nearline_node restriction. Then you'll compare to all files that have ever existed. This is probably fine.
- ... or compare to all files that currently exist, i.e. where any FileCopy with has_file=Y exists for that file.

ch_pipeline/processing/daily.py
```diff
@@ -407,29 +410,136 @@ def _load_hook(self):
...
+            di.ArchiveInst.name == "chimestack",
+            di.CorrFileInfo.start_time < end_time,
+            di.CorrFileInfo.finish_time >= start_time,
+            di.ArchiveFileCopy.node == nearline_node,  # that are on nearline
+            di.ArchiveFileCopy.has_file == "Y",
```
jrs65, line 492: The second suggestion above, with a bit of simplification to make it clearer that one set is a subset of the other
```suggestion
        # chimestack files available anywhere between start and end_time
        files_that_exist = archive_files.where(
            di.ArchiveInst.name
            == "chimestack",  # specifically looking for chimestack fiels
            di.CorrFileInfo.start_time
            < end_time,  # which contain data that includes start time and end time
            di.CorrFileInfo.finish_time >= start_time,
            di.ArchiveFileCopy.has_file == "Y",
        )

        # chimestack files available on online
        files_online = files_that_exist.where(
            di.ArchiveFileCopy.node == online_node,  # that are on online
        )
```

ch_pipeline/processing/daily.py
```diff
@@ -407,29 +410,136 @@ def _load_hook(self):
...
+
+        # For each intervals_weekly, find and add all CSDs, skipping every 7
+        for interval in self._intervals_weekly:
+            csd_i = csds_in_range(*interval, step=7)
             csd_set = set(csds)
             csds += [csd for csd in csd_i if csd not in csd_set]
 
-        tags = ["%i" % csd for csd in csds]
+        # only queue jobs for which all data is available in chime_online
+        tags = ["%i" % csd for csd in csds if self._available_files(csd, csd + 1)]
```
jrs65, line 542: I think this isn't the best way to do this. It's going to fire off 4 separate database queries for every CSD in the set specified by intervals. That's far too many queries.

A more efficient way to do this is to do one set of queries and return all files that exist anywhere, and all files that exist on online (you can probably restrict to between the min and max CSDs in your list), and then count the number of files in each category that overlap with the CSD. Only if those numbers are equal is the CSD "available".

jrs65 (changes requested):

ch_pipeline/processing/daily.py
anjakefala: > Also, these definitely should be defined in terms of their CSD and not their calendar date. Over the course of a year sidereal and calendar days shift by a whole year with respect to each other, so you will get off the expected schedule at some point.

I can definitely do this, I would love a little bit extra clarification! 

I chose dates instead of CSDs because there is a specific date until which we have calibration tables available for. Even with this, we still prefer CSD? Would we run into an issue where calendar days with unavailable calibration run?

jrs65: Ah, I see why you did this now.

What you have is probably fine, but as there is one more sidereal day in a year than calendar days if you pick the set of seven calendar start days, there will actually be eight sidereal days in there and the logic will miss one of them (and thus 1/8 of all days).

So you have a 1/50 chance of this breaking I think and thus probably best to specify them as CSDs.

As for your problem about calibration tables, just convert your end datetime into a CSD and use that.

ch_pipeline/processing/daily.py
```diff
@@ -479,29 +617,18 @@ def csds_in_range(start, end):
...
-        start_csd = ephemeris.unix_to_csd(ephemeris.ensure_unix(start))
-        start_csd = math.floor(start_csd)
-
-    if end.startswith("CSD"):
-        end_csd = int(end[3:])
-    else:
-        end_csd = ephemeris.unix_to_csd(ephemeris.ensure_unix(end))
-        end_csd = math.ceil(end_csd)
+    start_csd = datetime_to_csd(start)
+    end_csd = datetime_to_csd(end)
```
jrs65, line 631: When you factored this out, you've implicitly changed the math.ceil for end_csd to a math.floor. That might be okay, but I'm not sure.

jrs65 (changes requested):

ch_pipeline/processing/daily.py
anjakefala: Oh, I did not notice that! It makes sense why it had not been factored out then. I will remove that commit for now.

ch_pipeline/processing/daily.py
```diff
@@ -406,21 +408,110 @@ def _load_hook(self):
...
+            di.ArchiveFileCopy.select(
+                di.ArchiveFileCopy.node,
+                di.ArchiveFile,
+                di.ArchiveAcq,
+                di.ArchiveInst,
+                di.CorrFileInfo,
+            )
+            .join(di.ArchiveFile)
+            .join(di.ArchiveAcq)
+            .join(di.ArchiveInst)
```
ketiltrout, line 461: This join seems pointless and there's enough table merging going on already.  Yes, you can isolate chimestack files by looking at the instrument name, but that's also encoded in the acq name, which you've already included in your query.

I'd drop the `ArchiveInst` join completely and then change your `where()` criterion below from
```
di.ArchiveInst.name == "chimestack"
```
to
```
di.ArchiveAcq.name % "%_chimestack"
```
which will match all acquisitions whose name ends in "chimestack".

ketiltrout: If you _really_ don't want to rely on the acq name to select chimestack, then it'd be better to handle this like you do with the node selection.  Fetch the `chimestack` instrument record once:
```
chimestack_inst = di.ArchiveInst.get(name="chimestack")
```
and then use that in the where clause, again to avoid this unnecessary join:
```
where(
   # [...]
   di.ArchiveAcq.inst == chimestack_inst
   # [...]
   )
```
You might even be able to argue this is more efficient than pattern matching against acq name like I suggested above.

ch_pipeline/processing/daily.py
```diff
@@ -406,21 +408,110 @@ def _load_hook(self):
...
+        files_online = files_that_exist.where(
+            di.ArchiveFileCopy.node == online_node,  # that are online
+        )
+
+        # dict: {acqName/fileName: absolute path}
+        filenames_online = {
+            os.path.join(f.file.acq.name, f.file.name): os.path.join(
+                f.node.root, f.file.acq.name, f.file.name
+            )
+            for f in files_online
```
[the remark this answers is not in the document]
anjakefala: It magically happens! This code was tested, and works.

ch_pipeline/processing/daily.py
anjakefala: > You should fetch the start and end times for each file when you do the queries. (remember to specify them in the .select(...)).
> And then you just calculate which files overlap. Given that none of the files overlap with each other, and none of the days overlap with each other, there's probably an efficient way of computing the file-day overlaps that doesn't involve doing all pairs.

ch_pipeline/processing/daily.py
ketiltrout: Yeah, the second is better.  Mariadb's pattern matching is probably pretty good, but it's still going to be significantly heavier than an integer comparison.

ch_pipeline/processing/daily.py
```diff
@@ -429,10 +513,90 @@ def _available_tags(self):
...
+                start_time, end_time, filenames_that_exist
+            )
+
+            if (len(online) == len(exists)) and (len(online) != 0):
+                csds_available.append(csd)
+
+            # Since the csds, filenames_online, and filenames_that_exist are sorted by time,
+            # and non-overlapping
+            # as we check a csd, we are going to index further into the lists.
+            index_online = max(index_online - 3, 0)
```
jrs65, line 561: I still don't get the -3 in here. As you state the files are sorted and guaranteed non-overlapping so I don't think you need to roll the index back at all here.

ch_pipeline/processing/daily.py
```diff
@@ -406,21 +408,103 @@ def _load_hook(self):
...
+            < end_time,  # which contain data that includes start time and end time
+            di.CorrFileInfo.finish_time >= start_time,
+            di.ArchiveFileCopy.has_file == "Y",
+        )
+
+        files_online = files_that_exist.where(
+            di.ArchiveFileCopy.node == online_node,  # that are online
+        )
+
+        ChimestackFile = namedtuple("ChimestackFile", "name start_time finish_time")
```
jrs65, line 481: I think my comment about about peewee objects vs tuples vs named tuples was a bit unclear, but hopefully this makes more sense...

Using peewee objects gives you a nice well defined interface to the query results, with named attributes for accessing the information. However, you pay a bit of a price in object construction.

Using bare tuples is simpler and more efficient, at the price of there being no standard interface, and you can't use the returned tuples for the usual peewee magic queries (as they're not fancy objects). But if you know you've extracted all the information you need in that query then this is fine. I think you are in this regime here as these queries return all the info you need.

The named tuple approach here tries to combine both, but I think doesn't really bring through the advantages of either. It lacks the standard and expected interface defined by the data_index tables that the first approach would give, and it requires an extra conversion to a namedtuple in there which probably negates some of the efficiency and simplicity of the second approach.

I guess my recommendation for the future is to use one of the first two approaches. If you think it's reasonable to just use plain old tuples then do so, otherwise use peewee's objects directly. It's definitely not worth changing this now.

ch_pipeline/processing/daily.py
```diff
@@ -406,21 +408,103 @@ def _load_hook(self):
...
+        ]
+
+        # files_that_exist might contain the same file multiple files
+        # if it exists in multiple locations (nearline, online, gossec, etc)
+        # we only want to include it once
+        filenames_that_exist = []
+        for node, acq, name, start_time, finish_time in files_that_exist.tuples():
+            if not filenames_that_exist or filenames_that_exist[
+                -1
+            ].name != os.path.join(acq, name):
```
jrs65, line 495: I think what you are trying to do here is remove the duplicated entries, but I fear it's not very robust as it assumes that the same file names will appear consecutively and I don't see how that is actually guaranteed.

I think there's two options here:

- Re-write the query such that it is based on ArchiveFile's not ArchiveFileCopy's. That would guarantee uniquely returned files. However, that might be a bit of work.
- Use a Python container that will guarantee uniqueness (e.g. a set), and dump anything other than the start and end times from the query. This would should be easy.
```python
ChimestackFile = namedtuple("ChimestackFile", "start_time finish_time")
filenames_that_exist = set(
    ChimestackFile(start_time, finish_time)
    for start_time, finish_time in files_online.tuples()
)
filesnames_that_exist = sorted(list(filenames_that_exist))
```

ch_pipeline/processing/daily.py
```diff
@@ -429,10 +513,90 @@ def _available_tags(self):
...
+    def _files_in_timespan(self, start, end, files):
+        """
+        Parameters
+        ----------
+        start : float
+            unix timestamp
+        end : float
+            unix timestamp
+        files : list of dict or namedtuple
+            dicts/namedtuples contain attributes `start_time` and `finish_time`
```
jrs65, line 578: I'm pretty sure this won't accept dictionaries despite what the docstring says.

ch_pipeline/processing/daily.py
```diff
@@ -429,10 +513,90 @@ def _available_tags(self):
...
+        csds_available = self._csds_available_data(
+            csds_sorted, filenames_online, filenames_that_exist
+        )
+        csds = [csd for csd in csds if csd in csds_available]
+
         tags = ["%i" % csd for csd in csds]
 
         return tags
 
+    def _csds_available_data(self, csds, filenames_online, filenames_that_exist):
```
jrs65, line 532: This routine will break if `csds` is not sorted. Obviously you feed in a sorted list and so it will be fine in this case, but I think it could be improved if either: the routine did the sort on the csds list itself so it's guaranteed; or at the very least it should make it very clear in the docstring it needs to be sorted.

ch_pipeline/processing/daily.py
```diff
@@ -429,10 +513,90 @@ def _available_tags(self):
...
+        """
+        Return the subset of csds in `csds` for whom all files are online.
+
+        `filenames_online` and `filenames_that_exist` are a list of namedtuples or dictionaries,
```
jrs65, line 536: This won't actually work with dictionaries.

jrs65 (changes requested):
Okay. A few small changes and comments. Otherwise this is looking great!

I'll be available all day to comment on changes so hopefully this will all be finished by the time we speak.

ch_pipeline/processing/daily.py
anjakefala: A file can contain more than one sidereal day! But, I guess if we are not interested in the exact number of files, then it does not matter. One file will be dropped from the `online` and `exists` list. Okay, I will remove it.

ch_pipeline/processing/daily.py
anjakefala: Hahahhaha I am so sorry, in VisiData we have an `AttrDict` structure which allows accessing dict entries like attributes. :joy: That is why we were so confused!

anjakefala: Good catch!

ch_pipeline/processing/daily.py
jrs65: Oh I see. I don't think dropping a file from the lists is okay.

I was confused about what you return from file_in_timespan. I think either you need to either return the last file that was in the span (and currently its the first file outside it), or subtract one (but only one, not three) from the index outside of that function.

ch_pipeline/processing/daily.py
anjakefala: > standard interface

This is a very good point against Namedtuples!

ch_pipeline/processing/daily.py
anjakefala: I made this change because `set()` was not working for me for some reason. I now see that it was not working because `set()`s are un-ordered, and it shifted the order. The `sorted()` should fix that. I will make the change, this is better.

anjakefala: Bonus: this is so much faster!

ch_pipeline/processing/daily.py
anjakefala: If this routine were to be legitimately used by other functions, it should probably sort everything that it gets. I was hesitent to do that, because everything it gets is sorted already, and for now I would rather include the stipulation in the docstrings.

ch_pipeline/processing/daily.py
anjakefala: We ended up switching to tuples!

jrs65 (approved):

PR #96, opened 2022-02-12
fix(CHIMEExternalBeam): fix typo in frequency interpolation

sjforeman:
Fixes a minor typo I found while I was adapting this code for general telescopes in driftscan

### DIFF
--- a/ch_pipeline/analysis/beam.py
+++ b/ch_pipeline/analysis/beam.py
@@ -528,7 +528,7 @@ class TransitResampler(task.SingleTask):
         lza = regrid.lanczos_forward_matrix(xgrid, x, a=self.lanczos_width).T
 
         y = np.matmul(ygrid, lza)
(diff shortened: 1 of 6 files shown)


jrs65:
Looks like a bad mistake. Is it likely to have effected anything?

sjforeman:
@jrs65 I don't think frequency interpolation of external beam models has been used for anything mission-critical - for example, Seth's beam model exists at the exact frequencies we need, so there is no interpolation. I don't know where else frequency interpolation might have been used, but presumably the bug would have been discovered already if it had mattered for anything...

tristpinsm (approved):

jrs65 (approved):

jrs65:
Okay great. I think people have played with this code over the years (I think Paula may have tried to use it), but nothing mission critical.

PR #97, opened 2022-02-18
fix(analysis.fringestop): Pass local array.

tristpinsm:
The old way of passing an `MPIarray` no longer works.

### DIFF
--- a/ch_pipeline/analysis/fringestop.py
+++ b/ch_pipeline/analysis/fringestop.py
@@ -73,7 +73,7 @@ class FringeStop(task.SingleTask):
 
         # Fringestop
         fs_vis = tools.fringestop_time(
-            tstream.vis,
+            tstream.vis[:],
             times=tstream.time,
             freq=freq,
             feeds=feeds,


anjakefala:
> The old way of passing an MPIarray no longer works.

I don't understand. =( Please add more details.

tristpinsm:
> > The old way of passing an MPIarray no longer works.
> 
> I don't understand. =( Please add more details.

Without pulling out the array explicitly the fringestop routine doesn't get the right shape.

jrs65:
The new way is definitely the correct way of doing it, although I don't understand what has changed to mean that it suddenly has started to break.

anjakefala:
> although I don't understand what has changed to mean that it suddenly has started to break.

The MPIArray work has not been merged yet, so the `local_shape` attribute should not have been affected. This is also weird to me.

tristpinsm:
In any case, it is broken and this fixes it.

tristpinsm:
actually this was already fixed in the decorrelation correction PR so I'll just go ahead and merge that.

PR #98, opened 2022-02-22
Updates to holography fringestop processing

tristpinsm:
I changed how the fringestop processing groups transits to make it easier to process observations as we go. Let me know you think processing a single observation at a time makes sense. I also added support for gated holography.


sareda26 (dismissed):
Hey Tristan, 

Yeah, this way of doing things makes sense to me.

PR #99, opened 2022-06-15
feat(containers): support HFB archive data

jrs65:
This is a draft implementation of a container type and reader for the HFB data. To run this you'll need to use radiocosmology/caput#198

Things that work:
- Reading HFB files directly using `HFBData.from_file` including axes selections.
- Time, frequency and beam selections via the Reader. Notably the time selections will take a range of UTC times to load.

Things that don't work, or need to be tweaked:
- Compression when saving out containers.
- Distributed reading of data (this requires propagating the distributed parameters from the `_dataset_spec` into the load routines).
- Is this the right place for this code?
- Maybe match the expected Reader API a little better.

This code can be used like the following:
```python
from ch_pipeline.core import containers

fname = "/project/rpp-chime/chime/chime_online/20211022T223938Z_chime_hfb/hfb_00010307_0000.h5"
fnglob = "/project/rpp-chime/chime/chime_online/20211022T223938Z_chime_hfb/hfb_000*.h5"

# Load a single file in ondisk mode (so no data is read immediately)
d = containers.HFBData.from_file(fname, ondisk=True, mode="r")

# Load a range of containers, but don't read any actual datasets (just the ringmaps)
da = containers.HFBData.from_mult_files(fnglob, ondisk=True, mode="r", datasets=())

# Use the reader class to extract a range of time by UTC timestamp and down select on
# both frequency and beam
r = containers.HFBReader(fnglob)
r.select_time_range(1634900000.0, 1634953600.0)
r.freq_sel = slice(4, 10)
r.beam_sel = slice(4, 10)
dr = r.read()
```

### DIFF
--- a/ch_pipeline/analysis/calibration.py
+++ b/ch_pipeline/analysis/calibration.py
@@ -17,10 +17,10 @@ from ch_util import fluxcat
 from ch_util import finder
 from ch_util import rfi
 
-from draco.core import task
+from draco.core import task, containers
 from draco.util import _fast_tools
 
-from ..core import containers
+from ..core import containers as ccontainers
 from ..core.dataquery import _DEFAULT_NODE_SPOOF
 
 
@@ -1007,7 +1007,7 @@ class TransitFit(task.SingleTask):
 
         Returns
         -------
-        fit : containers.TransitFitParams
+        fit : ccontainers.TransitFitParams
             Parameters of the model fit and their covariance.
         """
         # Ensure that we are distributed over frequency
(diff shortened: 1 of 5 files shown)


jrs65:
@ljgray note that the scope of this branch/PR has expanded a bunch. It is now more loosely defined as the bunch of changes I need to make the pipeline work with the new MPIArray code, and with the new thermal calibration.

ljgray (approved):

rikvl (approved):

PR #101, opened 2022-06-30
feat(client): Add "metrics" command to display pipeline info

ljgray:
Added "chp metrics" command group and "list" command.

### DIFF
--- a/ch_pipeline/processing/base.py
+++ b/ch_pipeline/processing/base.py
@@ -361,14 +361,11 @@ class ProcessingType(object):
 
         return pending
 
-    def completed(self):
-        """Find all jobs which have been run and finished
-        with status "FINISHED" or "CRASHED". This is slow.
+    def crashed(self) -> list:
+        """Find all jobs which have crashed.
 
         Returns
         -------
-        finished : list
-            Return the tags of all successful jobs.
         crashed : list
             Return the tags of all failed jobs.
         """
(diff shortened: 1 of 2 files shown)


jrs65:
> Method of getting finished and crashed jobs is slow (iterating through nested directories, reading output of STATUS file for each job). Should be fine, not sure how else to approach without modifying how caput flags running/finished/crashed jobs.

Interesting that this is slow. Have you run in through a profiler? Sometimes cedar's IO can be pretty slow, although I've often found this kind of operation surprisingly fast. I think chp's startup time is annoyingly slow because of some of the things it does, so I'm interested as to how the IO compares to that.

ch_pipeline/processing/base.py
```diff
@@ -361,6 +361,55 @@ def pending(self):
...
+        def _read_status(path: "Path") -> str:
+            file = path / "job" / "STATUS"
+            with file.open() as fh:
+                return fh.readline().strip()
+
+        # Recursively search the target directory
+        # fmt: off
+        entries = [
+            [path.name, _read_status(path)] 
+            for path in working_path.rglob("*")
```
jrs65, line 393:
```suggestion
            for path in working_path.glob("*") 
```
One thing that might be slowing is that the `rglob` would be letting it recurse far deeper than it needs to. I don't think you need to go any deeper than the first level.

Within the `crashed` directory you probably want to let it go one level deeper, but you'll again just want a normal glob rather than `rglob` I would guess.

ch_pipeline/processing/base.py
```diff
@@ -361,6 +361,55 @@ def pending(self):
 
         return pending
 
+    def completed(self):
```
jrs65, line 364: One thing about this is that all the `FINISHED` jobs should have been moved out of the working directory by this point, and to be honest, even if a job has FINISHED in its STATUS file, if it hasn't been moved out of the working directory something went wrong and it should be considered as if it crashed.

ch_pipeline/processing/base.py
```diff
@@ -361,6 +361,55 @@ def pending(self):
...
+        if crashed_path.exists():
+            # fmt: off
+            additional_crashed = [
+                path.name
+                for path in crashed_path.rglob("*")
+                if file_regex.match(path.name)
+            ]
+            # fmt: on
+            crashed += additional_crashed
+
```
jrs65, line 410: You may want to consider issues of double counting of jobs in there. For instance, a specific tag may have crashed several times, and thus it could end up appearing in the `crashed` directory (potentially several times under different sub directories) *and* in either the working directory (if it hasn't finished yet), or, in the final directory.

I'm not sure exactly how you want to count those. Should the same tag be able to appear multiple times in different lists?

ch_pipeline/processing/client.py
```diff
@@ -229,6 +235,32 @@ def generate(revision, number, max_number, submit, fairshare, user_fairshare):
...
+    pending = [job for job in available if job not in not_pending]
+
+    click.echo("Fairshare: {0}".format(fs[0]))
+    click.echo("User Fairshare: {0}".format(fs[1]))
+    click.echo("Available: {0}".format(len(available)))
+    click.echo("Pending: {0}".format(len(pending)))
+    click.echo("Waiting: {0}".format(len(waiting)))
+    click.echo("Running: {0}".format(len(running)))
+    click.echo("Successful: {0}".format(len(success) + len(ls)))
+    click.echo("Failed: {0}".format(len(failed)))
```
jrs65, line 261: Stylistically I would definitely advocate for f-strings in here. Shorter and cleaner IMO (and apparently faster too).

You might also want to consider if the output should be in some more structured format to making later parsing easier.

jrs65 (changes requested):
Sorry for the belated review. This is looking nice! I made a few comments of things to consider.

ch_pipeline/processing/base.py
```diff
@@ -361,6 +361,55 @@ def pending(self):
...
+        file_regex = re.compile("^%s$" % self.tag_pattern)
+
+        def _read_status(path: "Path") -> str:
+            file = path / "job" / "STATUS"
+            with file.open() as fh:
+                return fh.readline().strip()
+
+        # Recursively search the target directory
+        # fmt: off
```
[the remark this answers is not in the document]
ljgray: Fair. I guess it doesn't follow the 80 characters per line convention.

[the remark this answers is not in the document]
jrs65: Yeah, black has settled on 88. It's probably pretty reasonable. I always found 80 too narrow so used 100, but 88 seems a reasonable compromise.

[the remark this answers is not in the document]
ljgray: Right, had a look and it's because it wants to flatten a longish list comp. to a single line, which would end up being ~110 characters. Thoughts?

ch_pipeline/processing/client.py
ljgray: Fixed in b31ece6. My thought process for output format is that this should be more about user readability, while any formatting/parsing logic should be done elsewhere

ch_pipeline/processing/base.py
ljgray: Resolved

ljgray:
Most items should be resolved. It involves some internal calls, so ls and queued are both getting called multiple times. Still, I'd rather isolate each method to give a specific metric rather than grouping them all together.

ljgray:
Still have to test but at airport so will check later/tomorrow

jrs65 (dismissed):
I think the code all looks good now. I'll just approve now, but could you rebase and tidy up the commits before merging?

PR #103, opened 2022-07-12
fix(test_daily): change test_daily pipeline default time to 120 minut…

ljgray:
…es from 60 minutes.


ljgray:
test_daily pipeline used 60 minutes as a default time, pretty much never actually completed in 60 mins. Doubled the default time parameter so pipeline should complete with default config.

PR #104, opened 2022-07-13
Ljg/profiling

ljgray:
Adds a flag to chp item generate to enable profiling with cProfile. Also doubled the default run time of the test_daily pipeline as the previous default 60 minutes wasn't usually long enough.

### DIFF
--- a/ch_pipeline/processing/base.py
+++ b/ch_pipeline/processing/base.py
@@ -389,10 +389,10 @@ def queue_job(script, submit=True, profile=False):
             cmd = "caput-pipeline queue %s"
         else:
             cmd = "caput-pipeline queue --nosubmit %s"
-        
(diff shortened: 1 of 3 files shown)


jrs65:
I'm on the fence about this one. Mostly as I've always figured you should do profiling as part of your testing but not in production, so I don't necessarily see us doing this very often, and I also don't see if we're going to patch through one of the options why not all of them (i.e. psutil or pyinstrument).

FWIW, to do this before I've always just done a `--no-submit`, and then modified the jobscript by hand to do the profiling.

Anyway, let me mull this one over.

jrs65:
We might consider as an alternative just putting in the documentation how to do a `--no-submit` job and add the profiling arguments.

PR #105, opened 2022-07-20
feat(init): add lazy versioning

ljgray:
Basic implementation from [StackOverflow](https://stackoverflow.com/a/52359211/18850844) (as found by @jrs65)

Seems to work fine with some basic experimentation but should be tested a bit more.

### DIFF
--- a/ch_pipeline/__init__.py
+++ b/ch_pipeline/__init__.py
@@ -11,7 +11,15 @@ Submodules
     synthesis
 
 """
-from ._version import get_versions
 
-__version__ = get_versions()["version"]
-del get_versions
+
+def __getattr__(name: str):
+    """Custom handling for __version__ global variable."""
+    if name == "__version__":
+        # No need to re-get __version__ if it already exists
+        v = globals().get(name, None)
+        from ._version import get_versions
+
+        return v if v else get_versions()["version"]
+
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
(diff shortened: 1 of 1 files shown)


ljgray:
Todo: test run pipeline with these changes

ljgray:
Interestingly, this doesn't seem to speed up chp calls much. The top-level module import `import ch_pipeline` happens much faster, but `from ch_pipeline.processing import client` runs very slowly, even with lazy version handling added to all modules imported by `client`.

jrs65:
> Interestingly, this doesn't seem to speed up chp calls much. The top-level module import `import ch_pipeline` happens much faster, but `from ch_pipeline.processing import client` runs very slowly, even with lazy version handling added to all modules imported by `client`. Obviously importing `client` is importing several other modules, whereas the only time spent when importing `ch_pipeline` was in `versioneer`. Something else is likely causing the slow `chp` startup time.

I wonder what else it's doing. Can you pass it run it via `python -m cProfile` and see what it's doing?

ljgray:
> I wonder what else it's doing. Can you pass it run it via `python -m cProfile` and see what it's doing?

prfl5.prof in my scratch space. Seems like a ton of importlib calls. I haven't dug into it much more than that yet. Each relevant import (caput, chimdb.core, etc...) has a lazy_version branch.

ljgray:
Proposing to close and reject this PR as I don't think it adds anything of value, at least as far as I can tell.

ljgray:
Final comment - profiled the same import on my own machine. Total time ~0.37 secs. Number of calls is comparable to running on cedar, total time is just way faster.

PR #107, opened 2022-08-08
fix(dataquery): forcing a list should actually force a list

ljgray:
Not very important, but function _force_list checks for lists by checking for __iter__ attribute. Tuples, generators, numpy ndarrays, ... all have a __iter__ attribute. Probably will never be an issue, but this change will ensure that the return is actually a list.

### DIFF
--- a/ch_pipeline/core/dataquery.py
+++ b/ch_pipeline/core/dataquery.py
@@ -53,7 +53,7 @@ def _force_list(val) -> list:
     """Ensure configuration property is a list."""
     if val is None:
         return []
-    elif hasattr(val, "__iter__"):
+    elif hasattr(val, "__iter__") and not isinstance(val, str):
         return list(val)
     else:
         return [val]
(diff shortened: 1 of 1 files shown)


ch_pipeline/core/dataquery.py
```diff
@@ -49,12 +49,12 @@
 _DEFAULT_NODE_SPOOF = {"cedar_online": "/project/rpp-krs/chime/chime_online/"}
 
 
-def _force_list(val):
+def _force_list(val) -> list:
     """Ensure configuration property is a list."""
     if val is None:
         return []
     elif hasattr(val, "__iter__"):
```
jrs65, line 56:
```suggestion
    elif hasattr(val, "__iter__") and not isinstance(val, str):
```
This obviously isn't your problem, but I think this code predates our Python 3 transition (when `str` did not have an `__iter__` method), and on Python 3 it will turn a string into an array of the individual characters.

jrs65 (changes requested):

jrs65 (approved):

PR #108, opened 2022-08-12
fix(daily): Update pipeline config for ThresholdVisWeight changes.

ljgray:
ThresholdVisWeight filter was updated to return a RFIMask object rather
than doing the masking in place. A separate task to apply the filter was
added.

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -214,15 +214,14 @@ pipeline:
     # flagged regions
     - type: draco.analysis.flagging.ThresholdVisWeight
       in: sstream
-      out: sstream_threshold
+      out: weight_mask
       params:
-          relative_threshold: 0.5
+          relative_threshold: 0.7
 
-    - type: draco.analysis.flagging.RFIMask
-      in: sstream_threshold
+    # Apply the RFI mask. This will modify the data in place.
+    - type: draco.analysis.flagging.ApplyRFIMask
+      in: [sstream, weight_mask]
       out: sstream_mask
-      params:
-          stack_ind: 66
 
     # Make a map of the full dataset
     - type: draco.analysis.ringmapmaker.RingMapMaker


ch_pipeline/processing/daily.py
```diff
@@ -213,15 +213,14 @@
     # flagged regions
     - type: draco.analysis.flagging.ThresholdVisWeight
       in: sstream
-      out: sstream_threshold
+      out: weight_mask
       params:
-          relative_threshold: 0.5
```
jrs65, line 218: Is there any rationale to changing this threshold?

ljgray: With the change from the mean/median being per-baseline to per-frequency, the relative threshold can be more aggressive without cutting off all of the higher frequencies. A relative threshold of 0.5 let a lot of bad data through, although 0.9 might be a bit too aggressive. Depends on the behaviour that we want, but I figure its better to be safe than to let bad data throughy

jrs65: Ideally we'd have some tests where we see the effects on the data of tweaking that param. I think there's two competing effects in the changes that have been made:

- Median is now per frequency and so should be more representative of the actual level. That means upping the level makes sense.
- Median now excludes missing samples when it is being calculated. This was biasing the level calculation low, and so to maintain a similar level of excision we would need to *lower* the threshold, but we are tightening it.

Overall I'd guess that 0.9 is too high. The regridder is fairly successful when only a few samples are missing, but this would cut those regions out as they'd probably always lower the weights by 10%.

I guess maybe we should have a good look at the data to see what the effects are.

ljgray:
@jrs65 merging this with threshold set at 0.7

jrs65:
I just got very confused as the `draco.analysis.flagging.RFIMask` task was no longer being called in the config, but somehow this change ended up wiping it out. I'm going to create a PR to restore it in a working manner.

ljgray:
Oh, that's really odd, not sure what I did there...

PR #109, opened 2022-08-19
feat(galt_auto_cont): Add container and task for saving raw Galt autos. 

sareda26:
This PR creates a task for extracting the Galt autocorrelations from raw holography acquisitions, before regridding, and saving them into a new container for longterm storage without keeping raw acquisitions.

### DIFF
--- a/ch_pipeline/core/containers.py
+++ b/ch_pipeline/core/containers.py
@@ -802,6 +802,39 @@ class Photometry(ContainerBase):
         return self.index_map["source"]
 
 
+class GaltAutocorrelation(ContainerBase):
+    _axes = ("freq", "pol", "time")
+
+    _dataset_spec = {
+        "auto": {
+            "axes": ["freq", "pol", "time"],
+            "dtype": np.complex64,
+            "initialise": True,
+            "distributed": True,
+            "distributed_axis": "freq",
+        },
+        "weight": {
+           "axes": ["freq", "pol", "time"],
+           "dtype": np.float64,
+           "initialise": True,
+           "distributed": True,
+           "distributed_axis": "freq",
+        },
+    }
+
+    @property
+    def auto(self):
+        return self.datasets["auto"]
+
+    @property
+    def weight(self):
+        return self.datasets["weight"]
+
+    @property
+    def fpga_count(self):
+        return self.index_map["time"]["fpga_count"]
+
+
 def make_empty_corrdata(
     freq=None,
     input=None,
--- a/ch_pipeline/core/io.py
+++ b/ch_pipeline/core/io.py
@@ -40,7 +40,6 @@ from ch_util import andata
 
 from draco.core import task, io
 
-
 class LoadCorrDataFiles(task.SingleTask):
     """Load data from files passed into the setup routine.
 
@@ -473,3 +472,52 @@ class FilterExisting(task.MPILoggedTask):
         )
 
         return sorted(list(new_files))
+
+class SaveGaltAutoCorrelation(task.SingleTask):
+    """Extract the autocorrelations of the Galt telescope from a holography acquisition."""
+
+    # YY, YX, XX
+    _galt_prods = [2450, 2746, 3568]
+
+    def process(self, data):
+        """Extract the Galt autocorrelations and write them to disk.
+        Parameters
+        ----------
+        data: TimeStream
+            A TimeStream container holding a raw holography acquisition.
+        Returns
+        -------
+        autocorrelation: containers.GaltAutocorrelation
+            A GaltAutocorrelation container holding the extracted Galt autos
+            as a function of frequency, polarization product, and time.
+        """
+        from .containers import GaltAutocorrelation
+
+        # Redistribute over freq
+        data.redistribute("freq")
+
+        # Dereference beam and weight datasets
+        beam = data.vis[:].view(np.ndarray)
+        weight = data.weight[:].view(np.ndarray)
+
+        # Load only the data corresponding to the Galt inputs
+        galt_auto = beam[:, self._galt_prods, :]
+        galt_weight = weight[:, self._galt_prods, :]
+
+        # Initialize the auto container
+        autocorrelation = GaltAutocorrelation(
+            pol=np.array([b"YY", b"YX", b"XX"]),
+            attrs_from=data,
+            freq=data.freq,
+            time=data.index_map["time"],
+            comm=data.comm,
+            distributed=data.distributed,
+        )
+
+        # Redistribute output container over frequency
+        autocorrelation.redistribute("freq")
+
+        autocorrelation.auto[:] = galt_auto
+        autocorrelation.weight[:] = galt_weight
+
+        return autocorrelation


ch_pipeline/core/containers.py
```diff
@@ -802,6 +802,39 @@ def source(self):
         return self.index_map["source"]
 
 
+class GaltAutocorrelation(ContainerBase):
+    _axes = ("freq", "pol", "time")
```
tristpinsm, line 806: To support common axes like `freq` and `time`, there are base classes in `draco` that you should inherit from. `draco.core.containers.FreqContainer` and `draco.core.containers.TODContainer`.

ch_pipeline/core/io.py
```diff
@@ -473,3 +472,52 @@ def next(self, files):
         )
 
         return sorted(list(new_files))
+
+class SaveGaltAutoCorrelation(task.SingleTask):
+    """Extract the autocorrelations of the Galt telescope from a holography acquisition."""
+
+    # YY, YX, XX
+    _galt_prods = [2450, 2746, 3568]
```
tristpinsm, line 480: it would probably be preferable to determine these from the `index_map` rather than hard code them. I don't think these indices would have changed, but in principle they are allowed to.

ch_pipeline/core/io.py
```diff
@@ -473,3 +472,52 @@ def next(self, files):
         )
 
         return sorted(list(new_files))
+
+class SaveGaltAutoCorrelation(task.SingleTask):
```
tristpinsm, line 476: Maybe this is a bit nit-picky, but I would suggest renaming this to something like `ExtractGaltAutocorrelation`and moving it out of `io` (since it only performs io if you choose to save the output, same as any task). I'm not sure where would be the best place for it... Maybe `analysis.calibration` since I assume this is meant for noise source analysis. `analysis.beam` could also work.

ch_pipeline/core/io.py
```diff
@@ -473,3 +472,52 @@ def next(self, files):
...
+            A GaltAutocorrelation container holding the extracted Galt autos
+            as a function of frequency, polarization product, and time.
+        """
+        from .containers import GaltAutocorrelation
+
+        # Redistribute over freq
+        data.redistribute("freq")
+
+        # Dereference beam and weight datasets
+        beam = data.vis[:].view(np.ndarray)
```
tristpinsm, line 500: the preferred way to this now is to use `local_array`
```suggestion
        beam = data.vis[:].local_array
```

ch_pipeline/core/io.py
```diff
@@ -473,3 +472,52 @@ def next(self, files):
...
+            freq=data.freq,
+            time=data.index_map["time"],
+            comm=data.comm,
+            distributed=data.distributed,
+        )
+
+        # Redistribute output container over frequency
+        autocorrelation.redistribute("freq")
+
+        autocorrelation.auto[:] = galt_auto
```
tristpinsm, line 520: I'm not sure but I think this might trigger a warning with the new `mpiarray`... The alternative would be to use `local_array` again.
```suggestion
        autocorrelation.auto[:].local_array[:] = galt_auto
```

But if it works without generating warnings with the latest `caput` I guess it's fine.

tristpinsm (changes requested):

ch_pipeline/core/containers.py
```diff
@@ -802,6 +802,39 @@ def source(self):
...
+
+    _dataset_spec = {
+        "auto": {
+            "axes": ["freq", "pol", "time"],
+            "dtype": np.complex64,
```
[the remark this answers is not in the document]
sareda26: Agreed, but in principle we also want the cross-correlation of the two 26m polarisations in addition to 26m XX and YY, and that has a phase, so I set the container to be complex.

ch_pipeline/analysis/calibration.py
```diff
@@ -2364,3 +2365,65 @@ def _calculate_uv(freq, prod, inputmap):
...
+        # Get the product map and inputs
+        prodmap = data.prod
+        ina, inb = prodmap["input_a"], prodmap["input_b"]
+
+        # Locate the Galt autocorrelations and cross-pol correlation
+        flag_YY = np.where((ina == galt_inputs[0]) & (inb == galt_inputs[0]), 1, 0)
+        flag_YX = np.where((ina == galt_inputs[0]) & (inb == galt_inputs[1]), 1, 0)
+        flag_XX = np.where((ina == galt_inputs[1]) & (inb == galt_inputs[1]), 1, 0)
+
+        auto_flag = (flag_YY + flag_YX + flag_XX).astype(bool)
```
ssiegelx, line 2403: I think the way this is done, there is no guarantee that the products picked out by `auto_flag` will have the polarisation order that you define when creating the output container (although in practice it may end up working out based on the way the CHIME holography acquisitions are ordered).

I would probably construct the polarisation axis as follows:

```
pol = np.array([input_map[aa].pol + input_map[bb].pol for aa, bb in prodmap[auto_flag]])

File: ch_pipeline/analysis/calibration.py [hunk 8]
```diff
@@ -2364,3 +2365,65 @@ def _calculate_uv(freq, prod, inputmap):
...
+
+        # Initialize the auto container
+        autocorrelation = containers.GaltAutocorrelation(
+            pol=np.array([b"YY", b"YX", b"XX"]),
+            attrs_from=data,
+            freq=data.freq,
+            time=data.index_map["time"],
+            comm=data.comm,
+            distributed=data.distributed,
+        )
```
ssiegelx, line 2421: I think you want to use `freq=data.index_map["freq"]` here to retain both the centre and width of each frequency channel.

More generally, you can pass `axes_from=data` as a keyword and then only specify the axes that are not present in data or that you would like to overwrite (which in this case would be the `pol` axis).

ch_pipeline/analysis/calibration.py
```diff
@@ -2364,3 +2365,65 @@ def _calculate_uv(freq, prod, inputmap):
...
+
```
ssiegelx, line 2422: I'm wondering if instead of creating a separate `GaltAutocorrelation` container, we should use the existing `draco.core.containers.TimeStream` container with only 2 inputs.  I think that has everything we need except for a way to specify the polarisations.  The polarisations of the two inputs could be saved as an attribute and then for later tasks that use the autocorrelation data, you could derive the polarisation products as follows:
```
autocorrelation.attrs["pol"] = np.array([input_map[ii].pol for ii in galt_inputs])
pol_prod = [autocorrelation.attrs["pol"][aa] + autocorrelation.attrs["pol"][bb] for aa, bb in autocorrelation.prod]
``` 
The benefit of doing it this way is that we reduce the overall number of containers and we can apply other pipeline tasks that have been developed for the CHIME timestreams to the holography autocorrelation data without any modifications to those tasks.

ch_pipeline/core/containers.py
```diff
@@ -802,6 +802,39 @@ def source(self):
         return self.index_map["source"]
 
 
+class GaltAutocorrelation(FreqContainer, TODContainer):
+    _axes = ("freq", "pol", "time")
```
ssiegelx, line 806:
```suggestion
    _axes = ("pol",)
```
If you inherit from the base classes, then you no longer need to specify the axes covered by those classes.   A new instance will have all the unique axes contained in all subclasses.

ch_pipeline/core/io.py
```diff
@@ -473,3 +472,4 @@ def next(self, files):
         )
 
         return sorted(list(new_files))
+
```
ssiegelx, line 475: I think these new lines can be be removed so that this PR does not modify the `io` module.

ssiegelx (changes requested):
Hi Alex, looks great, just a few comments.  The biggest one concerns the container type that we use to store the autocorrelations.

PR #113, opened 2022-09-06
convert ch_pipeline ringmap to draco ringmap container 

Arnab-half-blood-prince:
Covert the daily ch_pipeline ringmap to draco ringmap. The daily ringmap of ch_pipeline has rms axis instead of weight, so we are changing rms to weight and storing it in a new draco ringmap container.

### DIFF
--- a/ch_pipeline/analysis/mapmaker.py
+++ b/ch_pipeline/analysis/mapmaker.py
@@ -68,6 +68,11 @@ class RingMapMaker(task.SingleTask):
         tel : TransitTelescope
         """
 
+        self.log.warning(
+            "This task is deprecated.  Please use "
+            "draco.analysis.ringmapmaker.RingMapMaker instead."
+        )
+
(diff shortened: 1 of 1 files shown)


ssiegelx (dismissed):
Thanks @Arnab-half-blood-prince

jrs65:
Is there any reason to still use the `ch_pipeline` ring map maker? I think the draco one does a better job overall and is more extensible for the future.

If not, maybe we should add a deprecation warning to it so people don't use it any more? I've already switched the daily pipeline over to the draco one for revision 5.

ssiegelx:
No, I don't think so.  Arnab has been investigating the daily ringmaps that were generated with rev 03.  The purpose of this task is to allow us to process those ringmaps without adding checks for an `rms` dataset everywhere.

I will add a deprecation warning to the RingMapMaker task in ch_pipeline.

ssiegelx (approved):

PR #114, opened 2022-09-06
feat(calibration): pipeline to flag narrowband gain errors

ssiegelx:


ssiegelx:
I still need to go through and switch to the new `MPIArray` functionality, but otherwise I think this is good to go.

I've included an example config file for generating the correction.  I have already run this for the majority of the archive, and will make the files containing the correction available in `/project/rpp-chime/chime/chime_processed`.

The `ch_pipeline.analysis.calibration.FlagNarrowbandGainError` task uses these files to mask the data.  It can be applied to either the TimeStreams during the daily pipeline or the SiderealStreams prior to averaging over sidereal days when generating the quarter stacks.

tristpinsm (dismissed):
This looks great! One question I have is why are the narrowband features identified on common-mode gain data now, rather than per-input and then looking at the average error over input?

ssiegelx:
My thinking here is that both the stacked visibility data and maps are sensitive to the gain averaged over some large number of inputs.  If we generate a separate mask for each input, then the maximum deviation that can be identified is determined by the uncertainty on the gain of a single input.

However, there could be narrowband features that are below the noise level of a single input, but are present in many inputs.  These features would probably appear in the stacked visibility data and maps.  The uncertainty on the common-mode gain is much smaller than the uncertainty on the gain for single inputs, so you could potentially identify these features using the common-mode.

The code can accommodate both methods, so we can try both and see what works better.

ssiegelx:
To evaluate the performance, I re-stack the first 50 sidereal days that went into the `rev_03` sidereal stack.  These days cover January to mid-May 2019.  I apply the flags derived from the narrowband gain errors (using the `FlagNarrowbandGainError` task with a threshold of 1e-3) to the individual days prior to stacking.  I then create an intercylinder ringmap (simple beamforming, no deconvolution) and finally calculate the delay spectrum over RA 115 to 245 deg (which has complete coverage for this set of days).

[delay_spec_performance_evalutation_sstack_201901_201905.pdf](https://github.com/chime-experiment/ch_pipeline/files/9550847/delay_spec_performance_evalutation_sstack_201901_201905.pdf)

The first page of the PDF file compares (for XX pol) the delay spectrum following this procedure to the delay spectrum if you do not apply the flags.  The second page is the ratio of the delay spectrum without flags to that with flags.  The third and fourth page are the same figures for the YY pol.

Applying the flags results in a significant reduction in the excess power at high delays for the lower band.  The XX cleans up quite nicely.  There are still some declinations with very clear foreground leakage at high delays.  I'm guessing this corresponds to the declinations of bright point sources.  It is unclear if this leakage is due to residual unflagged gain errors or some other effect.  I will play around a bit more with the threshold that is used to see if there is any improvement here.

I don't see any significant improvement in the upper band.

There is also a large improvement for the YY pol in the lower band.  However, even with the improvement, it still looks really bad.  I will check how similar the gain error flags are for XX pol and YY pol.  Maybe we need to apply a separate set of flags for each polaristion.  Note that I am not applying any additional frequency mask for the lower band.  For the upper band, I derived a mask by manually identifying corrupted frequencies in the foreground filtered, stacked ringmap, but have not done the same for the lower band.

tristpinsm:
Those plots of the lower band are very encouraging! They seem to suggest that the regular vertical lines are associated with the narrowband gain errors. I can go through and flag individual frequencies again in the lower band with the flagged data eventually. For now I can also provide you with the mask I am using currently. It's located at `/project/rpp-chime/tristpm/lya/masks/mask_pre_filter_uniform_visual_v4.h5` on `cedar`. I think anybody in `rpp-chime` should be able to access it.

tristpinsm (dismissed):

PR #115, opened 2022-09-07
Update RFIFilter to use draco container and ApplyRFIMask task

ljgray:


ljgray:
@jrs65 Curious about your opinion, is it worth eventually deprecating the original behaviour (with the ch_pipeline container and ApplyCorrInputMask task) in favour of the draco option, or is it potentially needed elsewhere? I don't see the point in having multiple possible outputs if it isn't necessary

jrs65:
Yeah. I think longer term we should look how much this functionality is really needed, and whether to deprecate/replace or not.

It does seem like in general applying per feed RFI masks makes sense, but I'd be resistant to actually doing it in the CHIME pipeline because of the assumptions we make elsewhere.

ch_pipeline/analysis/flagging.py
```diff
@@ -62,6 +63,7 @@ class RFIFilter(task.SingleTask):
     freq_width = config.Property(proptype=float, default=10.0)
     time_width = config.Property(proptype=float, default=420.0)
     threshold_mad = config.Property(proptype=float, default=6.0)
+    use_draco_container = config.Property(proptype=bool, default=False)
```
jrs65, line 66: Please document this parameter for the class.

jrs65 (approved):
Approved pending a docstring line for the new parameter.

PR #116, opened 2022-09-07
Update configs for rev05 run

jrs65:
- config(daily): restore refactored second round RFI excision
- config(daily): write large outputs in zarr format

### DIFF
--- a/ch_pipeline/processing/base.py
+++ b/ch_pipeline/processing/base.py
@@ -73,6 +73,7 @@ class ProcessingType(object):
                 self.default_params,
                 encoding="utf-8",
                 allow_unicode=True,
+                sort_keys=False,
                 stream=fh,
             )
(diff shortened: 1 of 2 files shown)


jrs65:
@ljgray I think this config with the last releases of all the branches should be what we want to run.

ljgray:
Just curious, what is the rationale for using the zarr format over h5? I'm just not overly familiar with it beyond knowing of its existence

ljgray (dismissed):

jrs65:
You can't do parallel IO on chunked HDF5 files. They claim you can, but in reality it crashes (you can see a long series of issues we've filed about it on their forums). Zarr does at least work for that use case.

jrs65:
- [ ] Should change the delay spectrum setting to assume a complex time domain.

ljgray (approved):

PR #119, opened 2022-09-16
config(daily): add precision truncation to pipeline output

jrs65:


ch_pipeline/processing/daily.py
```diff
@@ -364,56 +460,67 @@ class DailyProcessing(base.ProcessingType):
     default_params = {
         # Time range(s) to process
         "intervals": [
-            ## Priority days to reprocess
-            # Good ranges from rev_00
+            # Two short intervals either side of the caltime change, and one
+            # with the weird 4 baseline issue in order to have a few good test
+            # days at the very start
+            {"start": "CSD1878", "end": "CSD1879"},
+            {"start": "CSD1912", "end": "CSD1913"},
```
sjforeman, line 467: Can I request to look at `CSD1960` here as well? It's the most egregious example of a day with a corrupted frequency at a single baseline that wasn't masked in rev_03. My work in progress on modifying `ThresholdVisWeightBaseline` should catch it, but if that doesn't make it into rev_05, it would be good to see whether some other rev_05 change catches it.

sjforeman: In that case, I'd also like to request `CSD1983` be put up front. It also has vertical stripes at certain baselines in rev_03 (attached plot only shows 3-cyl baselines - notice the stripes towards the right-hand side), but I haven't easily been able to localize the cause to specific time samples or frequencies by visual inspection. I'm interested to see how it looks in rev_05.
Screen Shot 2022-09-27 at 11 05 48 PM

ch_pipeline/processing/daily.py
```diff
@@ -134,10 +141,25 @@
...
+      out: decorr_cyl_mask
+      params:
+        threshold: 5.0
+        max_frac_freq: 0.1
+
     # Average over redundant baselines across all cylinder pairs
     - type: draco.analysis.transform.CollateProducts
       requires: manager
-      in: tstream
+      in: tstream_bbm
```
sjforeman, line 162: Since the bad-baseline masking was moved down in the pipeline, this should be changed back:
```suggestion
      in: tstream
```

ch_pipeline/processing/daily.py
```diff
@@ -364,56 +516,76 @@ class DailyProcessing(base.ProcessingType):
...
         "intervals": [
-            ## Priority days to reprocess
-            # Good ranges from rev_00
+            # Two short intervals either side of the caltime change (1878,
+            # 3000), one with the weird 4 baseline issue (1912), one with a
+            # single baseline-freq issue (1960), and one with a decorrelated
+            # cylinder (2325) in order to have a few good test days at the very
+            # start
+            {"start": "CSD1878", "end": "CSD1879"},
+            {"start": "CSD1912", "end": "CSD1912"},
```
sjforeman, line 525: I thought you always need `"end"` = `"start"+1`, i.e.:
```suggestion
            {"start": "CSD1912", "end": "CSD1913"},
```
(Same comment for 1960 and 2325 below)

sjforeman (changes requested):
Just a few small things...

ch_pipeline/processing/daily.py
jrs65: Thanks. Actually I've fixed that one locally as I wasn't getting any output!

ch_pipeline/processing/daily.py
jrs65: For this CSD spec thing. I made the ends *inclusive*. https://github.com/chime-experiment/ch_pipeline/blob/master/ch_pipeline/processing/daily.py#L650

That may be confusing I guess. We can consider changing it in the future.

sjforeman: Ah, okay. It would reduce confusion if there was a consistent convention for these single-day intervals in this file

jrs65: Ah, yeah, I can see that being confusing, although actually I did intend those to be two day intervals!

ch_pipeline/processing/daily.py
jrs65: This should be in the config now.

ch_pipeline/processing/daily.py
jrs65: Added a comment to clarify that.

sjforeman (approved):

PR #120, opened 2022-09-21
feat(daily): Apply frequency mask before estimating delay spectrum.

ljgray:


ljgray:
@jrs65 Note: this saves the delay filter container, which could be turned off if we don't care, and has complex time domain set to true.

jrs65:
Thanks @ljgray I've applied these changes directly so I'm going to close this PR.

PR #122, opened 2022-09-22
feat(analysis.flagging.IterHPFMask): Task to generate frequency mask …

tristpinsm:
…using iterative HPF.

I've translated the code I was using to generate frequency masks from ringmaps using the iterative HPF method that Seth developed to identify narrow-band-gain features. I've made the task such that it should work on other containers/datasets/axes, but more experimentation would be required to figure out where it could be useful.


ch_pipeline/analysis/flagging.py
```diff
@@ -1874,3 +1874,109 @@ def process(self, timestream):
...
+        rms_ind = axes.index(self.rms_axis)
+        freq_ind = axes.index("freq")
+        nf, nr = dv.shape[freq_ind], dv.shape[rms_ind]
+
+        # compute RMS and collapse remaining axes
+        rms = np.mean(np.sqrt(np.abs(dv) ** 2 * (dw != 0.0)), axis=rms_ind)
+        freq_ind = freq_ind if rms_ind > freq_ind else freq_ind - 1
+        rms = np.moveaxis(rms, freq_ind, -1).reshape(-1, nf)
+
+        # set initial frequecy mask according to missing data
```
jrs65, line 1955:
```suggestion
        # set initial frequency mask according to missing data
```

ch_pipeline/analysis/flagging.py
```diff
@@ -1874,3 +1874,109 @@ def process(self, timestream):
...
+
+        if self.dataset not in data.datasets:
+            raise ValueError(f"Container {type(data)} has no '{self.dataset}' dataset.")
+        axes = list(data[self.dataset].attrs["axis"])
+        if self.rms_axis not in axes:
+            raise ValueError(
+                f"Dataset '{self.dataset}' has no '{self.rms_axis}' axis'."
+            )
+
+        # figure out what is the largest remaning axis
```
jrs65, line 1933:
```suggestion
        # figure out what is the largest remaining axis
```

ch_pipeline/analysis/flagging.py
```diff
@@ -1874,3 +1874,109 @@ def process(self, timestream):
...
+        rem_axes = [a for a in axes if a not in ["freq", self.rms_axis]]
+        rem_ind = [axes.index(a) for a in rem_axes]
+        dist_axis = rem_axes[np.argmax([data[self.dataset].shape[i] for i in rem_ind])]
```
jrs65, line 1936: Maybe more Pythonic, though maybe less clear? Not really required, I was just having fun.
```suggestion
        rem_axes = [(data[self.dataset].shape[ii], a) for ii, a in enumerate(axes) if a not in ["freq", self.rms_axis]]
        _, dist_axis = max(rem_axes)
```

jrs65 (changes requested):

ch_pipeline/analysis/flagging.py
tristpinsm: 👍

PR #123, opened 2022-09-27
feat(io): Load HFB data with selections

rikvl:
This is ready for review.

Changes `core.io.LoadDataFiles` to load HFB data.

Selections in e.g. beam number or frequency index can be achieved as documented in `draco.core.io.BaseLoadFiles`.

### DIFF
--- a/ch_pipeline/core/io.py
+++ b/ch_pipeline/core/io.py
@@ -273,10 +273,10 @@ class LoadDataFiles(task.SingleTask):
         rd.select_time_range(time_range[0], time_range[1])
 
         # Select frequency range
-        rd.freq_sel(self.freq_sel)
+        rd.freq_sel = self.freq_sel
 
         # Select beams
-        rd.beam_sel(self.beam_sel)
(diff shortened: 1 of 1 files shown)


ch_pipeline/core/io.py
```diff
@@ -35,6 +35,7 @@
 
 from caput import pipeline
 from caput import config
+from caput import mpiutil
```
jrs65, line 38: I'm not quite sure why you need to add this import?

jrs65 (changes requested):
It's looking great. Just one question about an import.

ch_pipeline/core/io.py
rikvl: Right, that import is not needed for this PR.

My editor flagged a few lines in this file because `mpiutil` is missing, so it seems that perhaps the import _is_ needed, but that probably shouldn't be part of this PR.

I removed it.

jrs65 (approved):

PR #124, opened 2022-09-28
feat(flagging): create task for masking decorrelated cylinders

ssiegelx:
Creates a `MaskDecorrelatedCylinder` task that can be run in the daily pipeline prior to averaging over cylinder pair to identify decorrelated cylinders.  Outputs an RFIMask that when applied to the visibility data will mask any time and frequency where a decorrelated cylinder occured.

Provides [examples/test_decorr_cyl.yaml](https://github.com/chime-experiment/ch_pipeline/compare/ss/decorrelated_cyl?expand=1#diff-bb39f099ab51a8c4870537d5a946aa5820eb9403e5c1dc0f991618b8d60d8e1c) as an example of how to run the task.

The identification is more effective if the task is provided the map between frequency bin and FPGA motherboard slot.  I have generated this map for all past data based on the `fpga-master` configuration files and saved it to `/project/rpp-chime/chime/chime_processed/freq_map/20180902_20220927_freq_map.h5`.  This file would need to be appended to whenever the frequency remapping is updated in the real-time pipeline.  Ideally there would be a way to get the map automatically.

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -1705,6 +1705,27 @@ class MaskCHIMEMisc(task.SingleTask):
 class MaskDecorrelatedCylinder(task.SingleTask):
     """Identify and mask frequencies and times where a cylinder decorrelated.
 
+    If the error rate is high on a backplane link in the second stage shuffle
+    of the F-engine corner turn, then on rare occassions (few times per year)
+    the data streams being handled by that FPGA can become misaligned or
+    desynchronized with the rest of the data streams.  The 512 correlator inputs
+    on the cylinder corresponding to that pair of FPGA crates will have negligible
+    correlation with all other inputs for the 64 frequencies received by that
+    FPGA during the second stage shuffle.  This will persist until the data streams
+    are re-synchronized with a correlator restart.
+
+    This task identifies times and frequencies affected by these misalignment events
+    by examining, for each cylinder, the ratio of 1-cylinder-separation, co-polar
+    visibilities acquired by redundant baselines that do not contain the cylinder
+    to those that do contain the cylinder.  This ratio will be close to 1 under
+    normal operations since the baselines are largely redundant, however it will
+    become very large after a cylinder becomes misaligned since there is no
+    correlation and the denominator drops to near zero.
+
+    Additionally, if provided the mapping between frequency channel and FPGA,
+    this task will ensure that when there is evidence of a decorrelated cylinder,
+    all 64 frequencies handled by the problematic FPGA are masked.
+
     Parameters
     ----------
     threshold: int
@@ -1764,10 +1785,8 @@ class MaskDecorrelatedCylinder(task.SingleTask):
         pos = tools.get_feed_positions(inputmap)
 
         # Get cylinder each feed is on
-        cyl = np.array(
-            [chr(63 + inp.cyl) if tools.is_chime(inp) else "0" for inp in inputmap]
-        )
-        ucyl = np.unique(cyl[cyl != "0"])
+        cyl = np.array([inp.cyl if tools.is_chime(inp) else -1 for inp in inputmap])
+        ucyl = np.unique(cyl[cyl >= 0])
         ncyl = ucyl.size
 
         # Make sure that none of our typical products reference non-array feeds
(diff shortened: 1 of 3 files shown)


jrs65:
> Ideally there would be a way to get the map automatically.

For recent enough data (I think May 2020 onwards) this can be be pulled out of the dataset states, via the `f_engine_frequency_map` state type. However, I think *all* of the data before that has the 256 spacing map, so we could switch entirely to that method when we've implemented the lookup.

ch_pipeline/analysis/flagging.py
```diff
@@ -1702,6 +1702,348 @@ def process(self, ss):
...
+        t = data.prod[stack_new["prod"]]
+
+        prodstack = t.copy()
+        conj = stack_new["conjugate"]
+        prodstack["input_a"] = np.where(conj, t["input_b"], t["input_a"])
+        prodstack["input_b"] = np.where(conj, t["input_a"], t["input_b"])
+
+        # Calculate baseline distance and polarisation pair
+        index_a = prodstack["input_a"]
+        index_b = prodstack["input_b"]
```
jrs65, line 1798: I guess you can't use `data.prodstack` in here because pre-collation it occasionally gives you a pair of feeds where one doesn't actually map to a real CHIME input?

ch_pipeline/analysis/flagging.py
```diff
@@ -1702,6 +1702,348 @@ def process(self, ss):
...
+
+        # Get positions of feeds
+        pos = tools.get_feed_positions(inputmap)
+
+        # Get cylinder each feed is on
+        cyl = np.array(
+            [chr(63 + inp.cyl) if tools.is_chime(inp) else "0" for inp in inputmap]
+        )
+        ucyl = np.unique(cyl[cyl != "0"])
+        ncyl = ucyl.size
```
jrs65, line 1771: I think working with char's add a bunch of unnecessary complexity (especially as I don't think this task actually exposes any of the mapping to the user) and it makes some parts this PR harder to follow than they need to be. I would suggest just leave everything as ints if that can be made to work:
```suggestion
        # Get cylinder each feed is on (2 to 5 -> cyl_A to cyl_D)
        cyl = np.array([inp.cyl) if tools.is_chime(inp) else -1 for inp in inputmap])
        ucyl = np.unique(cyl[cyl < 0])
        ncyl = ucyl.size
```
I think there's one spot later down where a comparison needs to be changed. I'll attempt to add a suggestion there.

ch_pipeline/analysis/flagging.py
```diff
@@ -1702,6 +1702,348 @@ def process(self, ss):
...
+            if not bflag[bb]:
+                continue
+
+            pi = ico[isort[bnd[bb] : bnd[bb + 1]]]
+            pindex[cc] = valid_stack[pi]
+
+            flag_with[:, cc, :] = np.char.equal(
+                ucyl[:, np.newaxis], cyl[index_a[pi]][np.newaxis, :]
+            ) | np.char.equal(ucyl[:, np.newaxis], cyl[index_b[pi]][np.newaxis, :])
+
```
jrs65, line 1849: If just using ints, I think the `np.char.equal` just needs to become `==`
```suggestion
            flag_with[:, cc, :] = (
                (ucyl[:, np.newaxis] == cyl[index_a[pi]][np.newaxis, :]) 
                | (ucyl[:, np.newaxis] == cyl[index_b[pi]][np.newaxis, :])
            )
```

ch_pipeline/analysis/flagging.py
```diff
@@ -1702,6 +1702,348 @@ def process(self, ss):
         return ss
 
 
+class MaskDecorrelatedCylinder(task.SingleTask):
+    """Identify and mask frequencies and times where a cylinder decorrelated.
+
```
jrs65, line 1707: I think it might be useful for others to just have a few sentences describing the general algorithm in here. It is well commented throughout, but it would help someone coming in with no prior knowledge orient themselves a little more.

jrs65 (approved):
This looks great. I've made a few optional comments in here, but none are really critical.

ch_pipeline/analysis/flagging.py
ssiegelx: Yeah exactly.  I have to first use `tools.redefine_stack_index_map` to generate a stack index map that only references CHIME inputs and then copy over the code from `prodstack` get the products and perform the conjugation.

ch_pipeline/analysis/flagging.py
ssiegelx: Good point, there is no need to use chars in this task.

ch_pipeline/analysis/flagging.py
ssiegelx: I've added comments that describe these decorrelated cylinder events and the algorithm used to identify them.

ssiegelx:
> For recent enough data (I think May 2020 onwards) this can be be pulled out of the dataset states, via the `f_engine_frequency_map` state type. However, I think _all_ of the data before that has the 256 spacing map, so we could switch entirely to that method when we've implemented the lookup.

Ah, great.  I'll try this method for looking up the frequency map and compare to what I am getting from the FPGA configuration files.  If that makes sense then I'll update the task to use that method.  However, I don't think it needs to hold up merging this version and using this task in the daily pipeline.

jrs65 (approved):

ch_pipeline/analysis/flagging.py
ssiegelx: Yeah I think that would be nice.  Are you envisioning a separate pipeline task or doing this in `andata` / `LoadCorrDataFiles`?

It does require an input map queried from the layout database.  Perhaps an alternative that does not require the layout database would be to use products whose inputs are not always flagged based on the `input_flags`.

PR #125, opened 2022-09-30
Fixes for MaskDecorrelatedCylinder

jrs65:
- fix(MaskDecorrelatedCylinder): crash when given a subset of frequencies
- fix(MaskDecorrelatedCylinder): avoid NaN warning if entire sample is invalid


ssiegelx (approved):
Thanks!

PR #127, opened 2022-10-03
feat(client): option to select user when getting slurm jobs

### LINKED ISSUE #126: "Waiting" and "Running" tags not being found
Calling `revision.queued()`  does not seem to parse running or waiting jobs correctly and shows 0 for both.

ljgray:
`chp` commands for checking pending and running jobs would only get jobs being run by the current user. Jobs being run by a special account, such as the chime account (which is used for daily pipeline jobs) are visible to all users but could not be fetched by `chp` commands. This change adds a user option when finding slurm jobs, and adds a `-u` option when calling `chp metrics`. This also fixes #126

### DIFF
--- a/ch_pipeline/processing/client.py
+++ b/ch_pipeline/processing/client.py
@@ -231,14 +231,21 @@ def generate(revision, number, max_number, submit, fairshare, user_fairshare):
 
 @item.command("metrics")
 @click.argument("revision", type=PREV)
-def metrics_list(revision):
+@click.option(
+    "-u",
+    "--user",
+    type=str,
+    default=None,
+    help="User to check jobs for.",
+)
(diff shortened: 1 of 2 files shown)


ch_pipeline/processing/client.py
```diff
@@ -231,14 +231,21 @@ def generate(revision, number, max_number, submit, fairshare, user_fairshare):
 
 @item.command("metrics")
 @click.argument("revision", type=PREV)
-def metrics_list(revision):
+@click.option(
+    "-u",
+    "--user",
+    type=str,
+    default=None,
```
jrs65, line 238: Maybe set the default to `chime` in here.

jrs65 (dismissed):

PR #131, opened 2022-10-12
fix(containers): access to flags data via property

### LINKED ISSUE #130: Flags datasets not accessible from container properties
The `HFBData` container has a `weight` property that should return the `flags/hfb_weight` dataset, but I haven't been able to use this property (it gives a `KeyError`):
https://github.com/chime-experiment/ch_pipeline/blob/4aa51579c21f5191feec0e566d0b47d92753153e/ch_pipeline/core/containers.py#L962-L965
The weight data are accessible if I change line 965 to `return self._data["flags/hfb_weight"]`

The same seems to apply to `CHIMETimeStream.frac_lost`:
https://github.com/chime-experiment/ch_pipeline/blob/4aa51579c21f5191feec0e566d0b47d92753153e/ch_pipeline/core/containers.py#L1026-L1029

I guess this is related to the way `datasets` is defined in `draco.core.ContainerBase`, which skips `flags` (and the datasets inside) because `flags` is a group:
https://github.com/radiocosmology/draco/blob/90c994a421184be9b2e812606821ee645933ea63/draco/core/containers.py#L377-L394
```
    @property
    def datasets(self):
        """Return the datasets in this container.

        Do not try to add a new dataset by assigning to an item of this
        property. Use `create_dataset` instead.

        Returns
        -------
        datasets : read only dictionary
            Entries are :mod:`caput.memh5` datasets.

        """
        out = {}
        for name, value in self._data.items():
            if not memh5.is_group(value):
                out[name] = value
        return memh5.ro_dict(out)
```

Not sure if I'm making some basic mistake in how to deal with these classes, or if this is a bug and, if so, which part should be fixed (the container properties or `ContainerBase.datasets`).

rikvl:
Fixes #130 (the container properties `HFBData.weight` and `CHIMETimeStream.frac_lost` were not working).

This commit simply removes the `.datasets` from `self.datasets["flags/hfb_weight"]` and `self.datasets["flags/frac_lost"]`.

Still not sure whether this is the proper way to do this, but it seems to work.

### DIFF
--- a/ch_pipeline/core/containers.py
+++ b/ch_pipeline/core/containers.py
@@ -962,7 +962,7 @@ class HFBData(RawContainer, FreqContainer):
     @property
     def weight(self) -> memh5.MemDataset:
         """The inverse variance weight dataset."""
-        return self.datasets["flags/hfb_weight"]
+        return self["flags/hfb_weight"]
 
 
 class CHIMETimeStream(TimeStream, RawContainer):
@@ -1026,7 +1026,7 @@ class CHIMETimeStream(TimeStream, RawContainer):
     @property
     def frac_lost(self):
         """Get the input flags dataset."""
-        return self.datasets["flags/frac_lost"]
+        return self["flags/frac_lost"]
 
     @property
     def flags(self):


jrs65:
This looks good to me.

I'd forgotten the structure of the code. The `.datasets` attribute is actually part of `draco.core.containers.ContainerBase`, I think to match the API in `andata`. The code explicitly doesn't allow returning subgroups. What you've changed it to is probably the best way to work with this, although I'm starting to questions what the behaviour of the `.datasets` property should be.

jrs65 (approved):

PR #133, opened 2022-10-14
HFB RFI flagging

rikvl:
So far this includes:
- the new container `HFBRFIMask` for masks that flag RFI in HFB data,
- the new task `HFBRNTMask` to create these masks following @arashcosmology's `sensitivity_metric` method, averaging over beams.

Some questions:
1. Is it okay to (for now) leave out the `beam` axis in the mask container? Currently we make a selection of beams in the `LoadDataFiles` task upstream.
2. Is there some preference of which mask convention we should use? I.e., True = good data, False = bad data; or the other way round?
3. The next step would be to create an `ApplyHFBMask` task. In `ApplyCorrInputMask` the mask is applied by giving flagged data a weight of zero. Should we follow that approach?


rikvl:
Re the questions I had:
1. The `beam` axis is left out of the mask container. The beam selection is added to the mask container's attributes.
2. In the end, the mask convention used here is: True = bad data, False = good data.
3. An `ApplyHFBMask` was added. I tried to keep this fairly general (e.g., it should also work for masks that include a beam axis, or are missing other axes).

This is now ready for review.

It might be good to wait with merging this PR until after PR #134 is merged: if that is approved I can move these tasks to the HFB subpackage.

ch_pipeline/hfb/flagging.py
```diff
@@ -0,0 +1,145 @@
...
+        # Extract data and weight arrays, averaging over beams
+        data = np.mean(stream.hfb[:], axis=2)
+        weight = np.mean(stream.weight[:], axis=2)
+
+        # Number of samples per data point in the HFB data:
+        # N_S = Delta nu * Delta t, where Delta nu is the frequency resolution
+        # (390.625 kHz / 128) and Delta t the integration time (10 s)
+        # NOTE: Should this be extracted from the container attributes? In that
+        # case this information needs to be added to the attributes upstream.
+        N_SAMP = 390625.0 / 128.0 * 10.0
```
jrs65, line 53: Yes. This should definitely be pulled out of the container data. You should be able to determine the channel width from the freq index_map, and probably something similar for the subfreq.

ch_pipeline/hfb/flagging.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Ideal radiometer equation
+        radiometer = data**2 / N_SAMP
```
jrs65, line 56: I believe we also store a `frac_lost` dataset. That's important to use in case we have lost data upstream, and so the sample does not contain as much integration time as expected.

ch_pipeline/hfb/flagging.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Radiometer noise test: the sensitivity metric would be unity for
+        # an ideal radiometer, it would be higher for data with RFI
+        sensitivity_metric = 2.0 / (radiometer * weight)
+
+        # Boolean mask idicating data that are contaminated by RFI
```
jrs65, line 62:
```suggestion
        # Boolean mask indicating data that are contaminated by RFI
```

ch_pipeline/hfb/flagging.py
```diff
@@ -0,0 +1,145 @@
...
+                slc[ww] = None
+
+        # Extract mask
+        flag = mask.mask[:].astype(stream.weight.dtype)
+
+        # Convert mask from caput.mpiarray.MPIArray to regular numpy.ndarray
+        flag = np.array(flag)
+
+        # Expand mask to same dimension as weight array
+        flag = flag[tuple(slc)]
```
jrs65, line 127: I don't think all of this is strictly necessary, i.e. I don't think you need to upcast, and I also don't think you need to transform the array into a regular numpy array (and if you do, you should just do `flag.local_array`).

You might need to be a little careful with the MPIArray vs numpy array aspects when printing out the fraction flagged.

ch_pipeline/hfb/flagging.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Log how much data we're masking
+        self.log.info(
+            "%0.2f percent of data will be masked."
+            % (100.0 * np.sum(flag) / float(flag.size),)
+        )
```
jrs65, line 133: To do this with an MPIArray, probably something like this works
```suggestion
        # Log how much data we're masking
        frac = flag.mean().allreduce()
        self.log.info(f"{100 * frac:.2f} percent of data will be masked.")
```

ch_pipeline/hfb/flagging.py
```diff
@@ -0,0 +1,145 @@
...
+import numpy as np
+
+from caput import config
+
+from draco.core import task
+
+from containers import HFBRFIMask
+
+
+class HFBRNTMask(task.SingleTask):
```
jrs65, line 12: I think the name of the task should be more descriptive.

jrs65 (changes requested):

ch_pipeline/hfb/flagging.py
rikvl: Done.

ch_pipeline/hfb/flagging.py
rikvl: Added.

ch_pipeline/hfb/flagging.py
rikvl: Fixed.

ch_pipeline/hfb/flagging.py
rikvl: How about `HFBRadiometerNoiseTestRFIMask`?

rikvl: Converged on `HFBRadiometerRFIFlagging`

ch_pipeline/hfb/flagging.py
rikvl: Added a `TODO` comment in the code.

ch_pipeline/hfb/flagging.py
rikvl: Not applicable for now, since we left the numpy ndarray.

jrs65 (approved):

PR #134, opened 2022-10-18
Create HFB subpackage

rikvl:
In the near future, HFB-related tasks will require `beam_model` (i.e., [CHIMEFRB/beam-model](https://github.com/CHIMEFRB/beam-model)) to obtain FRB beam positions and sensitivities.

This PR adds `beam_model` as an optional dependency using `extras_require` in `setup.py`:
https://github.com/chime-experiment/ch_pipeline/blob/9754d4b00669e10058306d0f5278c7da4aaef164/setup.py#L9-L11

It also moves the existing HFB container and reader into a new subpackage called `hfb`. Any new HFB containers and tasks should (as much as possible) be added in this subpackage. When `beam_model` is needed, it can then be imported inside the `hfb` subpackage, but it should _not_ be imported elsewhere.


ch_pipeline/hfb/containers.py
```diff
@@ -0,0 +1,58 @@
+"""HFB-related containers
```
jrs65, line 1:
```suggestion
"""HFB containers
```

ch_pipeline/hfb/io.py
```diff
@@ -0,0 +1,99 @@
+"""HFB-related tasks
```
jrs65, line 1: This module does not contain tasks.

jrs65 (changes requested):

ch_pipeline/hfb/containers.py
rikvl: Done.

ch_pipeline/hfb/io.py
rikvl: Moved `HFBReader` back into the `hfb.containers` module and removed `hfb.io` for now (the way you originally set it up in `core.containers`). A future PR (#139) adds `hfb.io` again, putting `HFBDataLoader` in there. That `HFBDataLoader` is a task, so we can make the module docstring, for example,
```
"""HFB tasks related to reader and writing data
"""
```

jrs65 (approved):

PR #135, opened 2022-10-18
Merge solar beam code

wulfda02:
Given how stale the `dw/sunstop` branch had become, it was easier to start with a fresh branch.  All I've done is branch from master, overwrite `anaysis.solar` with the version from `dw/sunstop`, and copy the `SunTransit` container in `core.containers` from that same branch.

### DIFF
--- a/ch_pipeline/analysis/solar.py
+++ b/ch_pipeline/analysis/solar.py
@@ -2,7 +2,7 @@
 Tasks for analysis of the radio sun
 
 Tasks for analysis of the radio sun.  Includes grouping individual files
-into a solar day; solar calibration; and sun excision from sidereal stream.
+into a solar day; solar calibration; solar beamforming; and solar excision.
 
 Tasks
 =====
@@ -11,10 +11,10 @@ Tasks
     :toctree: generated/
 
     SolarGrouper
-    SolarCalibration
-    SolarClean
-    SolarCleanProject
+    SolarCalibrationN2
+    SolarCleanN2
     SolarBeamform
+    SolarClean
 
 Usage
 =====
@@ -22,14 +22,15 @@ Usage
 
 """
 
-from datetime import datetime
-import numpy as np
+import datetime
+import pytz
 
-from caput import config
-from caput import mpiutil
-from caput import time as ctime
-from ch_util import andata, ephemeris, tools, cal_utils
+import numpy as np
+import scipy.constants
+
+from caput import config, mpiutil, tod
 from draco.core import task
+from ch_util import ephemeris, tools, cal_utils
 
 from ..core import containers
 
@@ -47,15 +48,10 @@ def unix_to_localtime(unix_time):
     dt : :class:`datetime.datetime`
 
     """
-    import pytz
 
-    utc_time = pytz.utc.localize(datetime.utcfromtimestamp(unix_time))
+    utc_time = pytz.utc.localize(datetime.datetime.utcfromtimestamp(unix_time))
 
     return utc_time.astimezone(pytz.timezone("Canada/Pacific"))
-
-
-def _correct_phase_wrap(phi):
-    return ((phi + np.pi) % (2.0 * np.pi)) - np.pi
 
 
 def sun_coord(unix_time, deg=True):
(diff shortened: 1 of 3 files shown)


ssiegelx (changes requested):

ssiegelx:
Hi Dallas, I started working on a review, but then realized that we could get a significant speed up by refactoring `SolarBeamform`. 

Since the solar analysis only processes the y < 10 m baselines for a single cylinder and polarisation pair, the way the code is written means that 99.8% of the elements of the visibility matrix are being set to zero and then summed over at each iteration.  I modified the task so that it first downselects the visibilities to the desired baselines before beamforming.  It also now accumulates the different polarisations and cylinder pairs in a single pass (so the task doesn't have to be called multiple times with different values of cyl_id).

It now takes about 20 minutes to process a full day (all frequencies) with 2 nodes.  This is entirely set by the time it takes to read the files, whereas previously I think we would have spent about 25% / 75% of the time on reading / beamforming.

I went ahead and made the changes suggested in my review and some others.  I think the notable changes that I made to the `SolarBeamform` task are:

* Use `scipy.constants.c` for speed of light instead of 3e8.

* Use the `tools.redefine_stack_index_map` method to ensure that we are not using non-CHIME feeds in the reference products for unique baselines.  The previous version of the code was masking these baselines, but that data is good its just been labeled incorrectly.

* Change how the cross-polar data is being handled.  I explicitly find all YX products and conjugate them to transform to XY.  The polarisation axis out the output container is then `[XX, XY, YY]`.

* Change the output container from a `TimeStream` / `SiderealStream` to a newly created `FormedBeamTime` container.  This is similar to the `FormedBeamHA` container that is output by the `draco` beamformer task but with the following changes:

    - It has a time axis instead of an HA axis.
    - It uses a complex number to describe the beamformed visibilities.  I think saving the imaginary component for XX/YY is useful to check if we are beamforming correctly and if there is any significant relative phase due to feed-to-feed variations in the primary beam.  It also means that we can just save the cross-polar data as a single XY entry instead of reXY and imXY.
    - The different cylinders are saved as different elements of the `object_id` axis.

Also, I changed the names of some of the tasks to indicate that they require N2 data, so people don't get the impression that they can be run on the stacked visibilities.

Finally, I added an `exclude_nighttime` option to `ch_util.finder.Finder` and `core.io.QueryDatabase` so we can query a list of files that contain only the data that we want to process.  The `finder` changes are in [this PR](https://github.com/chime-experiment/ch_util/pull/42).

Let me know if you have any issues with the changes, and also if the format of the output container works for you.  I have confirmed that I recover the output of the previous version of the code after accounting for the changes above.

ssiegelx:
Here is the config file that I am using to test:

```
cluster:
    directory:          /scratch/ssiegel/chime/test/solar_beamform/all_freq/2_16_3_largemem_commit

    venv:               /home/ssiegel/chime_202206/venv

    ## Cluster configuration (ignored unless on a cluster)
    name:               sunstop                 # Jobname
    nodes:              2                       # Number of nodes to run on
    time:               60                      # Time to run for (in minutes)
    ompnum:             3                       # Number of OpenMP threads
    pernode:            16
    mem:                192000M
    system:            'cedar'
    account:           'rpp-chime'

# Pipeline task configuration
pipeline:

  logging:
    root: DEBUG
    peewee: INFO
    matplotlib: INFO

  save_versions:
    - caput
    - ch_util
    - ch_pipeline
    - chimedb.core
    - chimedb.data_index
    - chimedb.dataflag
    - cora
    - draco
    - drift
    - numpy
    - scipy
    - h5py
    - mpi4py

  tasks:

    - type: draco.core.task.SetMPILogging
      params:
        level_rank0: INFO
        level_all: WARNING

    # Query the files acquired over three days in April 2020
    - type: ch_pipeline.core.dataquery.QueryDatabase
      out: filelist
      params:
        start_time: "20200418T132801Z"
        end_time:   "20200421T063606Z"
        accept_all_global_flags: Yes
        exclude_nighttime: Yes
        node_spoof:
          cedar_online: "/project/rpp-chime/chime/chime_online/"
        acqtype: "corr"
        instrument: "chimestack"

    # Load the individual data files
    - type: ch_pipeline.core.io.LoadCorrDataFiles
      requires: filelist
      out: tstream

    # Get the telescope layout that was active when this data file was recorded
    - type: ch_pipeline.core.dataquery.QueryInputs
      in: tstream
      out: inputmap
      params:
        cache: true

    - type: ch_pipeline.analysis.solar.SolarBeamform
      in: [tstream, inputmap]
      out: beam
      params:
        ymax: 10
        exclude_intercyl: Yes
        sep_cyl: Yes
        
    - type: ch_pipeline.analysis.solar.SolarGrouper
      in: beam
      out: beam_full_day
      params:
        save: Yes
        output_name: "solar_beam_{tag}.h5"
```

wulfda02:
Hi Seth, These changes are fine with me.  Thanks for taking the time to improve the code!

ssiegelx (approved):

PR #138, opened 2022-10-26
chp methods to categorize some common job crashes

ljgray:
This PR does three things:
- Add methods to `base.py` to check log files for some common reasons why a job may crash
- Add client methods to check crash categories from the command line
- Add deprecation warning to a redundant `pending` method in `base.ProcessingType`


ljgray:
@jrs65 This is still a wip but it seems to properly detect failures due to NaN/Inf values and timeouts

ch_pipeline/processing/base.py
```diff
@@ -157,11 +159,20 @@ def ls(self):
         if not base.exists():
             raise ValueError("Base path %s does not exist." % base)
 
-        file_regex = re.compile("^%s$" % self.tag_pattern)
+        file_regex = re.compile(f"^{self.tag_pattern}$")
 
-        entries = [path.name for path in base.glob("*") if file_regex.match(path.name)]
+        entries = [
```
jrs65, line 164: I think it would be good to move the scanning for `mtime`'s into the `time_sort` conditional. Stat-ing files can be pretty slow on cedar and so it's best to avoid it if you can.

ch_pipeline/processing/base.py
```diff
@@ -347,26 +358,71 @@ def generate(self, max=10, submit=True):
             Submit the jobs to the queue.
```
jrs65, line 358: Needs a docstring for the `user` parameter.

ch_pipeline/processing/base.py
```diff
@@ -347,26 +358,71 @@ def generate(self, max=10, submit=True):
...
         """Jobs available to run."""
+        warnings.warn(
+            "'pending' method is deprecated. Call 'status()['not_yet_submitted']'."
+        )
+        return self.status(user)["not_yet_submitted"]
 
-        waiting, running = self.queued(user)
-        not_pending = set(self.ls()) | set(waiting) | set(running)
-        pending = [job for job in self.available() if job not in not_pending]
+    def failed(self, user=None, time_sort=False) -> Dict[str, list]:
```
jrs65, line 377: If you're adding type annotations to a method (👍 👍), then I would do the parameters too. Also, I usually don't add the types to the docstrings so they're only added in one place. I think the docstring parsers understand that and user the annotations.

ch_pipeline/processing/base.py
```diff
@@ -347,26 +358,71 @@ def generate(self, max=10, submit=True):
...
+        """Categorize failed jobs.
+
+        Parameters
+        ----------
+        user : str
+            user to find running jobs for. If not provided, this
+            will default to the current user
```
jrs65, line 384: Doc for `time_sort`

ch_pipeline/processing/base.py
```diff
@@ -347,26 +358,71 @@ def generate(self, max=10, submit=True):
...
+            ],
+            # Match anything else
+            "unknown": [re.compile(r"(.*?)")],
+        }
+        # Return all tags which match the listed patterns
+        return classify_failed(self.workdir_path, failed_tags, patterns)
 
-    def status(self, user=None) -> dict:
-        """Find the status of running jobs. This duplicates
+    def status(self, user=None, time_sort=False) -> dict:
```
jrs65, line 424: Again, docstring and type annotations.

Also be consistent with `Dict` vs `dict` (requires 3.9+) for annotations.

ch_pipeline/processing/base.py
```diff
@@ -464,6 +545,55 @@ def queue_job(script, submit=True):
         os.system(cmd % fh.name)
 
 
+def classify_failed(
+    dir: Union[Path, str], tags: list, patterns: dict = {}
+) -> Dict[str, list]:
+    """Analyze the cause of crashed jobs.
+
+    Parameters
```
jrs65, line 553: Again, probably remove the types from the docstring.

ch_pipeline/processing/base.py
```diff
@@ -464,6 +545,55 @@ def queue_job(script, submit=True):
...
+        # to find errors but it could be increased
+        process = sp.Popen(
+            ["tail", "-n", "300", str(file)],
+            stdout=sp.PIPE,
+            shell=False,
+            universal_newlines=True,
+        )
+        stdout, _ = process.communicate()
+        # Clean up the input
+        stdout = stdout.encode("unicode-escape").decode()
```
jrs65, line 586: I don't think there's a real need to use a subprocess for this. You can do it all in Python.

Reading the last N lines is slightly tricky to do efficiently, but given that there's no good reason to use exactly N lines what I would do is read the last N*M bytes where M is your estimate of the average length of a line in bytes (maybe ~60 characters). So open the file, seek to the starting position with `f.seek(-N * M, os.SEEK_END)` and then `f.read()`.

ch_pipeline/processing/client.py
```diff
@@ -140,9 +140,10 @@ def create(type_):
 @click.argument("revision", type=PREV)
 @click.option("-l", "--long", is_flag=True)
 @click.option("-h", "--human", is_flag=True)
-def item_list(revision, long, human):
+@click.option("-t", "--time", is_flag=True)
+def item_list(revision, long, human, time):
     """List existing items within the REVISION (given as type:revision)."""
-    for tag in revision.ls():
+    for tag in revision.ls(time):
```
jrs65, line 146: The meaning of this option isn't clear. So at the least it should have a `help=` string. If you're feeling awesome you could add them for the options above too.

ch_pipeline/processing/base.py
```diff
@@ -1,8 +1,10 @@
 import re
+import warnings
 import yaml
 import os
 import subprocess as sp
 import tempfile
+from typing import Dict, Union, Tuple
 
 # TODO: Python 3 workaround
```
jrs65, line 9: Can remove this block while you're editing this file? Thanks!

jrs65 (changes requested):

jrs65:
This all looked great. Super useful. A few mostly minor comments, but then it'll be good to go.

ch_pipeline/processing/base.py
ljgray: Ah, good call. I thought I had done that, but I guess I just thought about it...

ch_pipeline/processing/base.py
ljgray: If you don't add a type to the docstring how do you typically format it? ex:
```python
        crashed : dict
            tags associated with each category
```
just becomes
```python
        crashed :
            tags associated with each category
```
??

ljgray:
Changes have been addressed, but I just want to clarify about the usage of built-in vs `typing` type annotations (i.e. using `list` vs `List`). Is the suggestion that we stick to only using annotations imported from `typing` for the sake of compatibility with python 3.8?

ch_pipeline/processing/base.py
jrs65: I think that the convention is to drop the colon too:
```
        crashed
            tags associated with each category
```

jrs65:
> Changes have been addressed, but I just want to clarify about the usage of built-in vs `typing` type annotations (i.e. using `list` vs `List`). Is the suggestion that we stick to only using annotations imported from `typing` for the sake of compatibility with python 3.8?

I'm okay with bumping ch_pipeline Python 3.9 if we want too. But we should explicitly bump the minimum version.

jrs65 (dismissed):
Approved. Just figure out what you want to do about a version bump.

PR #139, opened 2022-10-28
HFB data loader with selections

rikvl:
Allows selection of
- frequency indices based on physical frequency, and of
- beam numbers based on source declination.

The task works like this in the config file:
```
        -   type:       ch_pipeline.hfb.io.LoadHFBDataFiles
            requires:   qfilelist
            out:        tstream
            params:
                source_dec: -1.98729406393
                freq_physical: [467.336]
```
To select multiple frequency bins, multiple physical frequencies can be added to the list: the bins in which these frequencies fall will be added. Maybe it would be good to add a way to specify a range of physical frequencies to select. What do you think?

The task also accepts manual selection of indices, like this:
```
        -   type:       ch_pipeline.hfb.io.LoadHFBDataFiles
            requires:   qfilelist
            out:        tstream
            params:
                selections:
                    freq_range: [851, 853]
                    beam_range: [12, 1024, 256]
```

A combination also works, like this:
```
        -   type:       ch_pipeline.hfb.io.LoadHFBDataFiles
            requires:   qfilelist
            out:        tstream
            params:
                source_dec: -1.98729406393
                selections:
                    freq_range: [851, 853]
```

Perhaps look at PR #134 before this one.

### DIFF
--- a/ch_pipeline/hfb/io.py
+++ b/ch_pipeline/hfb/io.py
@@ -1,6 +1,7 @@
 """HFB tasks for reading and writing files
 """
 
+import os
 import gc
 import numpy as np
 
@@ -16,8 +17,8 @@ from beam_model.formed import FFTFormedActualBeamModel
 from .containers import HFBReader
 
 
-class LoadHFBDataFiles(io.BaseLoadFiles):
-    """Load CHIME HFB data from a file list passed into the setup routine.
+class BaseLoadHFBDataFiles(io.BaseLoadFiles):
+    """Base class for loading CHIME HFB data from files on disk into containers.
 
     Attributes
     ----------
@@ -48,29 +49,16 @@ class LoadHFBDataFiles(io.BaseLoadFiles):
     passed, all frequencies/beams are read.
     """
 
-    files = None
-
-    _file_ptr = 0
-
     source_dec = config.Property(proptype=float, default=None)
     beam_ew_include = config.Property(proptype=list, default=None)
     freq_phys_list = config.Property(proptype=list, default=[])
     freq_phys_range = config.Property(proptype=list, default=[])
 
-    def setup(self, files):
-        """Set the list of files to load; set up frequency and beam selection.
-
-        Parameters
-        ----------
-        files : list
-        """
-        if not isinstance(files, (list, tuple)):
-            raise RuntimeError("Argument must be list of files.")
-
-        self.files = files
+    def setup(self):
+        """Set up frequency and beam selection."""
 
         # Resolve any selections provided through the `selections` attribute
-        self._sel = self._resolve_sel()
+        super().setup()
 
         # Set up frequency selection.
         cfreq = np.linspace(800.0, 400.0, 1024, endpoint=False)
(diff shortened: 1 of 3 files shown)


jrs65:
@rikvl now the first PR is merged can you rebase the others onto master so I can review them easily.

ch_pipeline/hfb/io.py
```diff
@@ -0,0 +1,178 @@
...
+    files = None
+
+    _file_ptr = 0
+
+    source_dec = config.Property(proptype=float, default=None)
+    beam_ew_exclude = config.Property(proptype=list, default=[])
+    freq_phys_list = config.Property(proptype=list, default=[])
+    freq_phys_range = config.Property(proptype=list, default=[])
+
+    def setup(self, files):
```
jrs65, line 59: I think we'll also want to support loading files specified in the config file. You may just need to derive a new class from this like all the variants in the main io tasks.

jrs65 (changes requested):

ch_pipeline/hfb/io.py
```diff
@@ -0,0 +1,178 @@
...
+    Method 1 takes precedence over method 2. If no relevant attributes are
+    passed, all frequencies/beams are read.
+    """
+
+    files = None
+
+    _file_ptr = 0
+
+    source_dec = config.Property(proptype=float, default=None)
+    beam_ew_exclude = config.Property(proptype=list, default=[])
```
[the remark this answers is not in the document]
rikvl: Done.

ch_pipeline/hfb/io.py
rikvl: I now created `BaseLoadHFBDataFiles` and `LoadHFBDataFromFilelist`.

rikvl: It seems we cannot use the `_list_of_filelists`, `_list_or_glob`, and `_list_of_filegroups` from `draco.core.io` as they are: currently we pass along a `time_range` for each group of files to only load data around transit time.

The HFB data loading was build around the `HFBReader(caput.tod.Reader)`, and was based on `ch_pipeline.core.io.LoadDataFiles`, which seems to be a different way of loading data, coexisting with the stuff found in `draco.core.io`.

Not sure how to proceed here:
1. Remove `HFBReader` and remake the data loading, more like `draco.core.io`.
2. Keep `HFBReader` and keep things more the way they are, but add some way of loading files specified in the config file.

rikvl: Went with option 2 in the above comment.

The structure in `ch_pipeline.hfb.io` now follows that of `draco.core.io` as much as possible:
1. The loading of files is handled by the method `_load_filelist` in `hfb.io.BaseLoadFiles`, which calls `hfb.containers.HFBReader`.
2. To load files specified in the config file (potentially with time ranges): use `hfb.io.LoadFilesFromParams`, which expects `filegroups` and processes them by calling `_load_filelist` (example given in documentation).
3. To load files specified by `core.dataquery.QueryDatabase`: use `hfb.io.LoadFiles`, which converts the output of the `QueryDatabase` to `filegroups`, so they can be processed by `LoadFilesFromParams`.

jrs65 (approved):

PR #141, opened 2022-12-19
Reading a time stamp

hgan0:
Add a time stamp to `static_rfi_mask_hook`.
`frequency_mask` in `ch_util/rfi.py `will read this timestamp for CSD-dependent static RFI flagging.

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -189,7 +189,7 @@ class RFISensitivityMask(dflagging.RFISensitivityMask):
 
         return madtimes
 
-    def _static_rfi_mask_hook(self, timestamp, freq):
+    def _static_rfi_mask_hook(self, freq, timestamp=None):
         """Use the static CHIME RFI mask.
 
         Parameters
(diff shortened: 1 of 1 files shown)


ljgray:
Quick note - I've merged a fix for the issue causing the "install-build-docs" test to fail, if you rebase from master that test will pass.

ch_pipeline/analysis/flagging.py
```diff
@@ -189,11 +189,14 @@ def _combine_st_mad_hook(self, times):
...
 
-    def _static_rfi_mask_hook(self, freq):
+    def _static_rfi_mask_hook(self, freq, timestamp=None):
         """Use the static CHIME RFI mask.
 
         Parameters
         ----------
+        timestamp : float
+            Start observing time (in unix time)
+
```
ljgray, line 199: This parameter should be below "freq" so it matches the order in which they're provided to the function.

ljgray (changes requested):
There's just one small change to the code that needs to be made, otherwise the changes look fine.

The one other change that I'll ask is to clean up the commit history a bit by rebasing into a single commit (using `git rebase`). I'm happy to help out with this because it's a bit tricky to figure out at first.

There's some info [here ](https://github.com/chime-experiment/Pipeline/blob/master/CONTRIBUTING.md) about some guidelines for working with the CHIME code, git in particular. It goes over what the final commits should look like before being merged into the master branches. This just helps keep the commit history of master clean and readable.

ljgray (approved):

PR #142, opened 2022-12-20
feat(containers): add time property to RawContainer

rikvl:
Returning the centres of the time bins, rather than the lower edge that `ctime` gives.

### DIFF
--- a/ch_pipeline/core/containers.py
+++ b/ch_pipeline/core/containers.py
@@ -944,7 +944,7 @@ class RawContainer(TODContainer):
         return parent_name == "/" or self.group_name_allowed(parent_name)
 
     @cached_property
-    def time(self):
+    def time(self) -> np.ndarray:
         """The 'time' axis centres as Unix/POSIX time."""
 
         time = self._data["index_map"]["time"]["ctime"].copy()
--- a/ch_pipeline/core/containers.py
+++ b/ch_pipeline/core/containers.py
@@ -25,6 +25,7 @@ Tasks
 """
 import posixpath
 from typing import List, Optional, Union
+from functools import cached_property
 from draco.core.task import MPILoggedTask
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


ch_pipeline/core/containers.py
```diff
@@ -942,6 +942,15 @@ def dataset_name_allowed(self, name: str) -> bool:
         parent_name, name = posixpath.split(name)
         return parent_name == "/" or self.group_name_allowed(parent_name)
 
+    @property
```
jrs65, line 945: I would consider caching this one. Either with the built-in `cached_property` decorator, or with the one we have in `caput.cache` if we think we want to support versions earlier than Python 3.8.

ch_pipeline/core/containers.py
```diff
@@ -942,6 +942,15 @@ def dataset_name_allowed(self, name: str) -> bool:
...
+    def time(self):
```
jrs65, line 946: Add a type annotation to this one.
```suggestion
    def time(self) -> np.ndarray:
```

jrs65 (changes requested):
Thanks for doing this one. It's looking good, just a few quick suggestions.

ljgray:
Quick note - I've merged a fix for the issue causing the "install-build-docs" test to fail, if you rebase from master that test will pass.

ch_pipeline/core/containers.py
rikvl: Added.

ch_pipeline/core/containers.py
rikvl: Using the one from `caput` gave the following error while testing:
```
TypeError: unhashable type: 'HFBData'
```
I now used the one from `functools` and didn't see the error anymore.

jrs65 (approved):

ljgray:
I haven't had a chance to fully figure out what's going on, but this commit (or the change to `cached_property`) seems to have introduced a breaking change when trying to access the `time` property. The basic traceback is:
```python
File "/project/6003614/lgray/chime_pipeline_dev/code/caput/caput/memh5.py", line 1205, in __getitem__
time = self._data["index_map"]["time"]["ctime"].copy()

return self._data[obj]
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```
I'm working on identifying the root of the issue so I just wanted to add this note.

PR #143, opened 2023-01-13
fix(containers): replace some deprecated numpy types with built-ins

ljgray:
This fixes the issues causing the build-docs check to fail in recent PRs. These numpy types were just aliases for the built-ins, so nothing changes with the behaviour.


ketiltrout (approved):
Nice.

PR #144, opened 2023-01-24
Add config option to prioritize processing recent CSDs 

ljgray:
Adds a config option to check `n` recent days since the current CSD and process those days before others in the queue.

Also, fix a bug so the daily config can actually handle not being provided an end date for some CSD range. In this case, any data between `start` and the current CSD will be included in the queue. 

The idea here is that a daily pipeline revision can process the most recent days first as their data become available. Note that I've set the default config num days to zero, so it won't check for recent days by default.

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -1,10 +1,10 @@
-import os
 import math
 import datetime
 
 import chimedb.core as db
-from ch_util import ephemeris
 import chimedb.data_index as di
+from ch_util import ephemeris
+from caput.tools import unique_ordered
 
 from . import base
 
@@ -586,6 +586,8 @@ class DailyProcessing(base.ProcessingType):
         ],
         # Amount of padding each side of sidereal day to load
         "padding": 0.02,
+        # Number of recent days to prioritize in queue
+        "num_recent_days_first": 0,
         # Frequencies to process
         "freq": [0, 1024],
         # The beam transfers to use (need to have the same freq range as above)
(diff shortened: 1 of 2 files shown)


ch_pipeline/processing/daily.py
```diff
@@ -713,41 +716,38 @@ def _available_tags(self):
...
         csds_sorted = sorted(csds)
+
+        # grab the list of files that are online, and that exist anywhere,
+        # from the earliest csd to the latest
         filenames_online, filenames_that_exist = self._available_files(
             csds_sorted[0], csds_sorted[-1] + 1
         )
-
         # only queue jobs for which all data is available online in chime_online
         csds_available = self._csds_available_data(
```
sjforeman, line 733: I haven't worked much with sets in Python, so apologies if this is a basic question, but: in changing `_csds_available_data()` to return a set instead of a list, is there a danger that the ordering of `tags` could be unintentionally different from the ordering of `csds_sorted`, because the latter gets converted to a set within `_csds_available_data()`? Or does this not happen, or not matter?

sjforeman (commented):
Looks good to me! I just had one question about `_csds_available_data()`

ch_pipeline/processing/daily.py
ljgray: In theory, a set should still maintain the order that items were added to it. However, this shouldn't actually matter here since we're pulling the csd order from `csds` rather than `csds_available`, so the order of `csds_available` won't matter. The reason I changed it to a set is because the lookup for whether an item exists in a set is O(1) on average, whereas it's O(n) for a list, so there's a marginal performance boost. The order in `csds` will be based on the revconfig file, so that's what we want to keep.

sjforeman: Ah of course, I overlooked the line where `csds_available` is simply used as a filter to select from the (ordered) elements of `csds`. Thanks!

sjforeman (approved):

Issue #146, opened 2023-01-31
`RawContainer` time shift issues


ljgray:
There are a couple of underlying issues with changing the `.time` property of `RawContainer` to centre time samples:
1. The `.time` property is not perfectly compatible with `draco` containers - example, applying a `draco` RFIMask to a `CHIMETimeStream` fails because the `.time` properties do not line up. This is because the underlying `index_map` is not changed, so the first entry in the `index_map["time"]` array is different than the first entry returned by `.time`. I could fix this by making some modifications to `ch_pipeline` RFIMask and adding a slight wrapper around `draco` RFIMask to handle the times correctly, but there is the more underlying issue of this incompatibility when dealing with any time-series containers.
2. `CollateProducts` copies the time axis directly from the `.time` property, rather than the `index_map`. This results in:
    - An additional shift in the time axis such that it has actually been shifted by a full sample
    - A change in the way the `index_map["time"]` property is indexed, causing this implementation to fail. This is accounted for in the `draco` `.time` property implementation. 

All of these issues couldbe fixed by implementing a task to directly shift the index_map time, rather than adding the new `.time` property.

PR #147, opened 2023-02-01
Remove `CHIMETimeStream` `.time` property and replace with a task to apply shift

ljgray:
This fixes the issues described in #146. In summary, using the modified `.time` property results in potential incompatibilities with other tasks and `draco` containers which use the `.time` property. 

I'm open to thoughts about different ways to implement this, but with this task applied early in the daily pipeline all products will be fine. However, it would have to be applied to any processing or analysis that pulls directly from raw corr data rather than using daily pipeline products.

### DIFF
--- a/ch_pipeline/analysis/timing.py
+++ b/ch_pipeline/analysis/timing.py
@@ -381,3 +381,45 @@ class ConstructTimingCorrection(task.SingleTask):
 
         # Return timing correction
         return tcorr
+
+
+class CentreTimeSamples(task.SingleTask):
+    """Apply a shift to container time samples.
+
+    Parameters
+    ----------
+    method : str [center, value]
+        How to shift the time samples. If `center`, samples are shifted
+        by half the absolute median difference. If `value`, samples are
+        shifted by the value of `shift` property.
+    shift : float
+        Shift to apply to the time samples, if `method` is `value`
+    """
+
+    direction = config.enum(["forward", "back"], default="forward")
+
+    def process(self, cont):
+        """Apply a shift to the container `time` property.
+
+        Parameters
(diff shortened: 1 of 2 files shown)


ljgray:
This doesn't have to be the definitive solution to the problem, but it avoids the unintended effects of the property implementation.

jrs65:
I think there's a misunderstanding about the set of fixes that need to be in various places (including myself in there):

- The `RawContainer.time` property absolutely needs to be in there, its absence was the fundamental cause of the mislabelling (as `.time` should always return the centre, but the underlying timestreams have `.index_map["time"]` refer to the sample start).
- The remedy that we proposed was to run things through `ShiftRA` task. Given an equivalent 5s shift that *should* mostly undo the issues (although that is up for debate).

I think the case you're addressing is the case when you want to deal with time ordered data that has the wrong labelling. I think in this case there's a mix of states things can be in:
- Loading raw data via `andata.CorrData`, `.time` will have the correct centre (and always will have)
- Turning raw data into a `CHIMETimeStream`, `.time` will be wrong (start) on old versions of `ch_pipeline` and correct (centre) on newer ones.
- Loading data (e.g. RFIMasks) derived from `CHIMETimeStream`'s in rev06. These will probably have the wrong time.

However, I'm very confused about this last point. With my current understanding about how the timestamping worked, I don't see how the mask and timestream `.time` properties ever aligned (which is what is being checked in the `ApplyRFIMask` task), so there's something I'm missing. @ljgray let's go through all this later on this morning.

ljgray:
I think we have a better understanding of the problems now, so I'm going to close this PR to avoid bothering people (but leave the issue open).

PR #161, opened 2023-02-10
New chp type to run some common sidereal processing tasks

ljgray:
This PR introduces a new chp processing type, which runs the barebones daily pipeline steps to produce a delay spectrum from a sidereal stream. The type is linked to an existing daily processing revision, so the available tags are pulled from the tags completed by the linked day. The linked daily revision is by default the most recent, but this can be changed in the revconfig file after creation. 

The idea is that anyone using this type is going to want to be customizing the jobtemplate in order to do something different from the linked pipeline revision. This just provides the barebones template, which should reproduce the delay spectrum that was created by the daily revision, assuming the venv is using the same commits (this would have to be manually configured).

It also adds a chp type for generating the wavelet spectrum of a given day based on radiocosmology/draco#237

### DIFF
--- a/ch_pipeline/processing/client.py
+++ b/ch_pipeline/processing/client.py
@@ -8,7 +8,11 @@ click.disable_unicode_literals_warning = True
 
 
 # Map the type names to classes
-_typedict = {t.type_name: t for t in base.all_subclasses(base.ProcessingType)}
+_typedict = {
+    t.type_name: t
+    for t in base.all_subclasses(base.ProcessingType)
+    if hasattr(t, "type_name")
+}
 
 
 class PType(click.ParamType):
--- a/ch_pipeline/processing/client.py
+++ b/ch_pipeline/processing/client.py
@@ -2,7 +2,7 @@
 
 import click
 
-from . import base, beam, daily, fullstack, hfb, quarterstack  # noqa: F401
+from . import base, beam, daily, fullstack, hfb, quarterstack, reprocess  # noqa: F401
 
 click.disable_unicode_literals_warning = True
 
--- a/ch_pipeline/processing/reprocess.py
+++ b/ch_pipeline/processing/reprocess.py
@@ -1,0 +1,198 @@
+"""Processing type to reprocess daily data products."""
+
+from typing import ClassVar
+
+from caput.util.arraytools import unique_ordered
+
+from . import base, client, daily
+
+DEFAULT_SCRIPT = """
+# Cluster configuration
+cluster:
+  name: {jobname}
+  account: rpp-chime
+
+  directory: {dir}
+  temp_directory: {tempdir}
+
+  time: {time}
+  system: slurm
+  nodes: {nodes}
+  ompnum: {ompnum}
+  pernode: {pernode}
+  mem: 768000M
+
+  venv: {venv}
+  module_path: {modpath}
+  module_list: {modlist}
+
+# Pipeline task configuration
+pipeline:
+
+  logging:
+    root: DEBUG
+    peewee: INFO
+    matplotlib: INFO
+
+  save_versions:
+    - caput
+    - ch_util
+    - ch_pipeline
+    - chimedb.core
+    - chimedb.data_index
+    - chimedb.dataflag
+    - cora
+    - draco
+    - drift
+    - fluxcat
+    - numpy
+    - scipy
+    - h5py
+    - mpi4py
+
+  tasks:
+
+    - type: caput.pipeline.tasklib.base.SetMPILogging
+      params:
+        level_rank0: DEBUG
+        level_all: WARNING
+
+    - type: caput.pipeline.tasklib.debug.CheckMPIEnvironment
+      params:
+        timeout: 420
+
+    # Load the telescope model that we need for several steps
+    - type: draco.core.io.LoadProductManager
+      out: manager
+      params:
+        product_directory: "{product_path}"
+
+    # Load the sidereal stream we want to process
+    - type: caput.pipeline.tasklib.io.LoadFilesFromParams
+      out: sstream
+      params:
+        files: "{src_type_path}/{csd}/{file_match}*{csd}.*"
+"""
+
+
+class SiderealReprocessing(base.ProcessingType):
+    """Baseclass for a reprocessing type.
+
+    Reprocess from a product of a daily pipeline revision.
+
+    A revision of this processing type is directly linked to an existing revision of the Daily
+    processing type and pulls available files from there. By default, it is linked to the most
+    recent Daily Processing revision, and can be changed in the revconfig file after creation.
+    In theory, this will be used to quickly re-generate products from an already-processed
+    Daily type, with custom changes made to the config.
+
+    This is non-functional on its own and is designed to be subclassed.
+    """
+
+    type_name = "reprocess"
+    tag_pattern = r"\d+"
+
+    # Default processing parameters
+    default_params: ClassVar = {
+        # Which type of file to glob against
+        "file_match": "tstream",
+        # Just include all CSDS for now - user can specify narrower range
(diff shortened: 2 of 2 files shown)


sjforeman (commented):
This looks good to me overall, although I won't approve it because Richard should probably have the final say.

A few comments:

- Should we consider using the 82-day rev06 stack in BlendStack instead of the rev03 stack? Occasionally in rev06 validation, we've seen hints of artifacts (e.g. vertical lines in delay spectra) that may be coming from the rev03 stack instead of individual rev06 days themselves, and presumably the rev06 stack is cleaner overall.

- Should we consider another level of abstraction for `DelaySpectrumProcessing`? Most of the code in that class isn't specific to the precise reprocessing we want to do. Thus, we could consider bundling that code into a generic template (called `ReProcessing` or something), and then have `DelaySpectrumProcessing` as a subclass of that, only specifying the modifications to `default_params`, `default_script` etc. In the future, if we need to some other partial reprocessing of a revision, we'd only need to subclass `ReProcessing` and write the corresponding script.

ljgray:
> This looks good to me overall, although I won't approve it because Richard should probably have the final say.

Agreed

> A few comments:
> 
>     * Should we consider using the 82-day rev06 stack in BlendStack instead of the rev03 stack? Occasionally in rev06 validation, we've seen hints of artifacts (e.g. vertical lines in delay spectra) that may be coming from the rev03 stack instead of individual rev06 days themselves, and presumably the rev06 stack is cleaner overall.

This can definitely be the default, although that should perhaps be a change for the standard Daily config in rev_07 of the daily pipeline. For reprocessing daily:rev_06, I think we should continue to use the rev_03 stack so no additional changes are introduced other than the new static mask.

>     * Should we consider another level of abstraction for `DelaySpectrumProcessing`? Most of the code in that class isn't specific to the precise reprocessing we want to do. Thus, we could consider bundling that code into a generic template (called `ReProcessing` or something), and then have `DelaySpectrumProcessing` as a subclass of that, only specifying the modifications to `default_params`, `default_script` etc. In the future, if we need to some other partial reprocessing of a revision, we'd only need to subclass `ReProcessing` and write the corresponding script.

I like this idea, although I think we will have to put some thought into the implementation so that we don't end up with a bunch of different `default_scripts`. I'll make a quick change to this in a separate commit to experiment a bit.

ljgray:
I've also added a task to get the static rfi mask and broadcast to the shape of the incoming data, and updated the reprocessing config to reflect the thresholdvisweight refactor

jrs65 (changes requested):
This looks really nice. Even if we don't end up using it, having a ready-to-go implementation of the functionality will be great for when we do need to do something similar.

For instance I can imagine running Minori's bandpass averaging code over the processed days using all this which would be great.

ch_pipeline/analysis/flagging.py
```diff
@@ -139,6 +142,48 @@ def process(self, data) -> Union[containers.RFIMask, dcontainers.RFIMask]:
         return out
 
 
+class RFIStaticMask(task.SingleTask):
```
jrs65, line 145: This might be better with a more active name. `MaskStaticRFI` might be better. We haven't been very consistent before, but I think we maybe should get better on that front.

ch_pipeline/analysis/flagging.py
```diff
@@ -139,6 +142,48 @@ def process(self, data) -> Union[containers.RFIMask, dcontainers.RFIMask]:
...
+    Parameters
+    ----------
+    invert : bool, optional
+        Whether to invert the mask. Defaults to False.
+    """
+
+    invert = config.Property(proptype=bool, default=False)
+
+    def process(self, data) -> Union[dcontainers.RFIMask, dcontainers.SiderealRFIMask]:
+        """ """
```
jrs65, line 160: Empty docstring?

ch_pipeline/analysis/flagging.py
```diff
@@ -139,6 +142,48 @@ def process(self, data) -> Union[containers.RFIMask, dcontainers.RFIMask]:
...
+
+    Parameters
+    ----------
+    invert : bool, optional
+        Whether to invert the mask. Defaults to False.
+    """
+
+    invert = config.Property(proptype=bool, default=False)
+
+    def process(self, data) -> Union[dcontainers.RFIMask, dcontainers.SiderealRFIMask]:
```
jrs65, line 159:
```suggestion
    @overload
    def process(self, data: dcontainers.TimeStream) -> dcontainers.RFIMask:
        ...
    @overload
    def process(self, data: dcontainers.SiderealStream) -> dcontainers.SiderealRFIMask:
        ...
    def process(self, data):
```
For my own interest I was trying to figure out how to type this. It's a bit problematic because what you really want are intersection types (apparently), e.g. data must be a `SiderealContainer` and a `FreqContainer` but with no other constraints, but they don't really exist. I think the most likely things to apply this too are TimeStreams and SiderealStream's so this is about the best that can be done.

jrs65 (changes requested):

ch_pipeline/analysis/flagging.py
ljgray: Seems like there was some discussion about it [here](https://github.com/python/typing/issues/18) but it doesn't look like there are any plans to introduce an Intersection/Any type in the way that we would want here. So I agree that using `overload` is probably the best option at the moment

ch_pipeline/analysis/flagging.py
ljgray: Oops

ch_pipeline/analysis/flagging.py
```diff
@@ -117,7 +117,10 @@ def process(self, data) -> Union[containers.RFIMask, dcontainers.RFIMask]:
         # Create container to hold output
         if self.use_draco_container:
             # draco RFIMask container is not distributed
```
ljgray, line 119: I had a look into how to type this, and I can't really come up with a way to do it (since it isn't conditional on input type). It's maybe a suggestion that we should refactor somehow, but I'm not sure if that's really worth it for a fairly basic task

jrs65: Yeah. I think the best you can really do is have the output be the Union of all possible types.

For me the most useful thing is editor autocompletion, and I think all of those types have a pretty similar interface so it shouldn't really matter too much.

ch_pipeline/analysis/flagging.py
```diff
@@ -68,7 +68,9 @@ class RFIFilter(task.SingleTask):
     threshold_mad = config.Property(proptype=float, default=6.0)
     use_draco_container = config.Property(proptype=bool, default=False)
 
-    def process(self, data) -> Union[containers.RFIMask, dcontainers.RFIMask]:
+    def process(
+        self, data
```
sjforeman, line 72: Should we include a type hint here as well?
```suggestion
        self, data: Union[ch_util.andata.CorrData, dcontainers.TODContainer, dcontainers.SiderealContainer]
```
Even if not, we should update the docstring for the `data` parameter, below

sjforeman (commented):
Looks great! I had only one minor comment to add. All of my previous comments have been addressed, so I'll leave it to @jrs65 to provide a final approval if he's happy as well

ch_pipeline/analysis/flagging.py
ljgray: Good call, I'll make that change

ljgray:
@sjforeman I'm revisiting this because it's become relevant for some students. Most of the excess stuff is removed - this is basically now just a processing type to do something with data products from some daily revision, with the assumption that the user can specify the type of file they want to deal with (default is tstream) and that the user will expand the jobtemplate file to suit their needs

sjforeman (approved):

PR #162, opened 2023-02-11
HFB fix divide by zero in RFI flagging

rikvl:
Bug fix.

### DIFF
--- a/ch_pipeline/hfb/flagging.py
+++ b/ch_pipeline/hfb/flagging.py
@@ -5,6 +5,7 @@ import numpy as np
 from caput import config
 
 from draco.core import task
+from draco.util import tools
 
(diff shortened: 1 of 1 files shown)


ch_pipeline/hfb/flagging.py
```diff
@@ -5,6 +5,7 @@
 from caput import config
 
 from draco.core import task
+from draco.util import tools
```
ljgray, line 8:
```suggestion
from caput import tools
```

ljgray (changes requested):
The version of `invert_no_zero` in `draco.util` was deprecated a little while back and now it just throws a warning and calls the version in `caput.tools`, so you should just directly import `caput.tools`. Otherwise this is good.

PR #163, opened 2023-02-13
Frequency Map (1/2): Query database for frequency slot map

ljgray:
Query the database for the frequency map that was active when a dataset was collected. Lookup is done by dataset_id, so older datasets which do not have a dataset_id will just use the default mapping that we assume was active at the time. 

Changes:
1.  Add a task to query the database for the frequency map corresponding to a dataset_id.
2.  Add a container type to hold a single frequency map. This does not modify the existing FrequencyMap container for backwards compatibility.
3. Modify the `MaskDecorrelatedCylinder` task to use the new container
4. Modify the daily config to use the new task(s)

Requires chime-experiment/ch_util#50

### DIFF
--- a/ch_pipeline/core/dataquery.py
+++ b/ch_pipeline/core/dataquery.py
@@ -37,13 +37,17 @@ in a dataspec YAML file, and loaded using :class:`LoadDataspec`. Example:
                 -   start:  2014-07-28 11:00:00
                     end:    2014-07-31 00:00:00
 """
+from __future__ import annotations
 
 import os
 
+import numpy as np
+
 from caput import mpiutil, config, pipeline
 from draco.core import task
-from chimedb import data_index as di
-from ch_util import tools, ephemeris, finder, layout
+from chimedb import data_index as di, dataset as ds
+from ch_util import tools, ephemeris, finder, layout, andata
+from ch_pipeline.core import containers
 
 
 _DEFAULT_NODE_SPOOF = {"cedar_online": "/project/rpp-krs/chime/chime_online/"}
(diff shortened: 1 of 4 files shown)


ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+        Parameters
+        ----------
+        ts : andata.CorrData
+            Timestream container.
+
+        Returns
+        -------
+        freq_map : FrequencySlotMap
+            Frequency slot map container
+        """
```
jrs65, line 688:
```suggestion
    def next(self, ts: andata.CorrData) -> FrequencyMapSingle:
        """Generate an input description from the timestream passed in.

        Parameters
        ----------
        ts
            Timestream container.

        Returns
        -------
        freq_map
            Frequency slot map container
        """
```

ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+
+        # Fetch from the cache if we can
+        if self.cache and self._cached_inputs:
+            self.log.debug("Using cached inputs.")
+            return self._cached_inputs
+
+        stream_id = None
+        stream = None
+
+        if mpiutil.rank0:
```
jrs65, line 698:
```suggestion
        if self.comm.rank == 0:
```

ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+        cont.stream[:] = stream
+        cont.stream_id[:] = stream_id
+
+        # Save into the cache for the next iteration
+        if self.cache:
+            self.log.debug("Caching input.")
+            self._cached_inputs = cont
+
+        # Make sure all nodes have container before return
+        mpiutil.world.Barrier()
```
jrs65, line 761: I don't think this final barrier is actually necessary. They should be pretty synchronized by the broadcasts above.

ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+                    # Extract the stream_id and frequency map
+                    fmap_dict = list(fmaps.values())[0]["fmap"]
+
+                    stream_id = np.fromiter(fmap_dict.keys(), dtype=np.int64)
+                    fmap = np.array(list(fmap_dict.values()), dtype=np.int64)
+                    stream = tools.order_frequency_map_stream(fmap, stream_id)
+
+        # Broadcast to all ranks
+        stream_id = mpiutil.world.bcast(stream_id, root=0)
+        stream = mpiutil.world.bcast(stream, root=0)
```
jrs65, line 740:
```suggestion
        stream_id = self.comm.bcast(stream_id, root=0)
        stream = self.comm.bcast(stream, root=0)
```

ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+        stream_id = None
+        stream = None
+
+        if mpiutil.rank0:
+            layout.connect_database()
+
+            # Check to see if this timestream has a dataset id, and assign it
+            # if it does. If there is no dataset_id, we will just assume a
+            # default mapping
+            if (did_flag := ts.flags.get("dataset_id", None)) is not None:
```
jrs65, line 704:
```suggestion
            if (dsid_flag := ts.flags.get("dataset_id", None)) is not None:
```
We usually denote these as dsid/ds_id or similar. did is a little confusing when reading.

ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+                    for dsid in unique_dataset_ids:
+                        d = ds.Dataset.from_id(dsid)
+                        state = d.closest_ancestor_of_type(
+                            "f_engine_frequency_map"
+                        ).state
+                        fmaps[state.id] = state.data
+
+                    if len(fmaps.keys()) > 1:
+                        raise ValueError(
+                            "Multiple freqeuncy maps found for this dataset."
```
jrs65, line 728:
```suggestion
                            "Multiple frequency maps found for this dataset."
```

ch_pipeline/core/dataquery.py
```diff
@@ -655,6 +658,111 @@ def next(self, ts):
...
+                did_flag = memh5.ensure_unicode(did_flag[:])
+                # Get only unique dataset ids to reduce number of lookups. Ignore
+                # the null dataset at the 0th index
+                unique_dataset_ids = np.unique(did_flag)[1:]
+
+                if len(unique_dataset_ids) != 0:
+                    fmaps = {}
+                    # Cast each entry to a python string due to a bug with mySQL
+                    # typ conversion
+                    unique_dataset_ids = [str(dsid) for dsid in unique_dataset_ids]
```
jrs65, line 715: You might be able to use the routines here (https://github.com/chime-experiment/chimedb_dataset/blob/master/chimedb/dataset/utils.py#L76) to simplify this. Have a look, but I might not be worth it.

jrs65 (changes requested):

ch_pipeline/core/dataquery.py
ljgray: Made some tweaks based on this, but can't call it directly since I also need `state.data` and would have to do that lookup anyway

jrs65 (approved):

PR #169, opened 2023-02-18
HFB target list

rikvl:
For now, this is the list of targets in JSON format.

Machinery to use the list still needs to be added.


jrs65 (changes requested):
This looks good, but I think we figured it would be best to move this to `ch_util`.

rikvl:
Moved to `ch_util`.

PR #173, opened 2023-03-01
Dataset_id alternative 1 (1/1): Convert distributed `dataset_id` dataset to a type useable by h5py

ljgray:
Converts `dataset_id` to be stored with `S33` type, which can be handled by h5py. Also introduces a property to access the dataset_id dataset converted to numpy unicode.

### DIFF
--- a/ch_pipeline/analysis/calibration.py
+++ b/ch_pipeline/analysis/calibration.py
@@ -1803,8 +1803,8 @@ class ThermalCalibration(task.SingleTask):
         # results
         try:
             # First attempt to group all dataset_ids for all frequencies on
-            # rank=0 so it can do the full lookup
-            dataset_ids = data.flags["dataset_id"][:].gather(rank=0)
+            # rank=0 so it can do the full lookup.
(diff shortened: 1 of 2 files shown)


jrs65:
I think I don't understand the underlying issues with what is currently implemented. Can you give a bit of background?

ljgray:
> I think I don't understand the underlying issues with what is currently implemented. Can you give a bit of background?

This is to address the issue in radiocosmology/draco#213 (forgot to link in PR). Since `dataset_id` is distributed, it won't get converted from unicode to bytestring by `memh5` when a `CHIMETimeStream` is being saved. 

An alternative is just to make dataset_id a common dataset and have it get checked properly on save, which I'm not opposed to unless it has to be distributed for some reason.

ljgray:
Actually, I realize that I did test it as a common dataset, which runs into the issue of it having a `freq` axis. Since the `CHIMETimeStream` container is distributed, the `freq` axes won't match if `dataset`_id` is common, which is why I implemented it this way.

Yet another option is to handle this conversion directly in `andata` when the file is being read

PR #175, opened 2023-03-07
Dataset_id (2/2): Ensure CHIMETimeStream datasets are accurate and handle distributed `dataset_id`

ljgray:
This PR does some things:
1. Ensures that `CHIMETimeStream` containers will only have datasets set in its spec, and that those datasets will be of the correct type when inheriting from `andata.CorrData` containers. This only affects the `flags/frac_rfi` dataset, which isn't use via the draco containers interface as far as I can tell. If it is needed, then a dataset_spec entry should be added for it.
2. Access `dataset_id` in unicode via a property on `CHIMETimeStream`
3. Properly call `dataset_id` wherever it's used
4. Fix a bug where `CHIMETimeStream.flags` would throw an error if certain datasets weren't present

Requires chime-experiment/ch_util#53

### DIFF
--- a/ch_pipeline/analysis/calibration.py
+++ b/ch_pipeline/analysis/calibration.py
@@ -1804,7 +1804,7 @@ class ThermalCalibration(task.SingleTask):
         try:
             # First attempt to group all dataset_ids for all frequencies on
             # rank=0 so it can do the full lookup
-            dataset_ids = data.flags["dataset_id"][:].gather(rank=0)
+            dataset_ids = data.dataset_id[:].gather(rank=0)
 
             if self.comm.rank == 0:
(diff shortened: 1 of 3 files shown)


sjforeman (dismissed):
This looks good to me. https://github.com/chime-experiment/ch_util/pull/53 looks like its commits could use some cleaning up before merging, however

ljgray:
> This looks good to me. [chime-experiment/ch_util#53](https://github.com/chime-experiment/ch_util/pull/53) looks like its commits could use some cleaning up before merging, however

Sounds good. I cleaned up the commits in the other PR, so that should be good to go as well

jrs65 (approved):

PR #176, opened 2023-03-09
rev_07 config updates

ljgray:
Daily pipeline config updates for rev_07.

Generally, this includes:
- Masking of extremely large or small weights (>1e30, <1e-30)
- ~~Generating and saving (but not applying) the sensitivity-based RFI mask~~
- Producing an additional delay-filtered ringmap
- Zipping zarr files in a background process
- Masking 3 additional rfi bands before making the delay spectrum. These bands seem to come from the stack.
- Reducing total run time

### DIFF
--- a/ch_pipeline/core/dataquery.py
+++ b/ch_pipeline/core/dataquery.py
@@ -75,6 +75,8 @@ class QueryDatabase(task.MPILoggedTask):
     node_spoof : dictionary
         (default: {'cedar_online': '/project/rpp-krs/chime/chime_online/'} )
         host and directory in which to find data.
+    connection_attempts : int (default: 1)
+        number of times to try to establish a database connection before exiting
     start_time, end_time : string (default: None)
         start and end times to restrict the database search to
         can be in any format ensure_unix will support, including e.g.
@@ -126,6 +128,7 @@ class QueryDatabase(task.MPILoggedTask):
     return_intervals = config.Property(proptype=bool, default=False)
 
     node_spoof = config.Property(proptype=dict, default=_DEFAULT_NODE_SPOOF)
+    connection_attempts = config.Property(proptype=int, default=1)
 
     acqtype = config.Property(proptype=str, default="corr")
     instrument = config.Property(proptype=str, default=None)
@@ -175,7 +178,7 @@ class QueryDatabase(task.MPILoggedTask):
             if self.run_name:
                 return self.QueryRun()
 
-            layout.connect_database()
+            layout.connect_database(ntries=self.connection_attempts)
 
             f = finder.Finder(node_spoof=self.node_spoof)
 
@@ -293,10 +296,13 @@ class QueryRun(task.MPILoggedTask):
         Name of the `run` defined in the database.
     node_spoof : str, optional
         Node spoof argument. See documentation of :class:`ch_util.finder.Finder`.
+    connection_attempts : int (default: 1)
+        number of times to try to establish a database connection before exiting
     """
 
     run_name = config.Property(proptype=str)
     node_spoof = config.Property(proptype=dict, default=_DEFAULT_NODE_SPOOF)
+    connection_attempts = config.Property(proptype=int, default=1)
 
     def setup(self):
         """Fetch the files in the specified run.
@@ -313,7 +319,7 @@ class QueryRun(task.MPILoggedTask):
 
         # Query the database on rank=0 only, and broadcast to everywhere else
         if mpiutil.rank0:
-            layout.connect_database()
+            layout.connect_database(ntries=self.connection_attempts)
 
             cat_run = (
                 layout.global_flag_category.select()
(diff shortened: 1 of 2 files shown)


sjforeman:
I have a feature request for rev_07: adding the following tasks after the DelayFilter,
```
    # Make an intercylinder ringmap from the delay-filtered visibilities,
    - type: draco.analysis.ringmapmaker.RingMapMaker
      requires: manager
      in: sstream_dfilter
      out: ringmap_int_hpf
      params:
        single_beam: true
        weight: natural
        exclude_intracyl: true
        include_auto: false
        
    # Select 10 frequencies from the delay-filtered map that are useful for validation
    - type: draco.analysis.transform.SelectFreq
      in: ringmap_int_hpf
      out: ringmap_int_hpf_sel
      params:
        channel_index: [65, 250, 325, 399, 470, 605, 730, 830, 950, 990]

    # Precision truncate and write out the chunked filtered ringmap.
    # Don't truncate the map itself, to preserve low-amplitude pixels.
    - type: draco.core.io.Truncate
      in: ringmap_int_sel
      params:
        dataset:
          map: No
          weight: 1.0e-5
        compression:
          map:
            chunks: [1, 1, 32, 512, 512]
          weight:
            chunks: [1, 32, 512, 512]
          dirty_beam:
            chunks: [1, 1, 32, 512, 512]
        save: true
        output_name: "ringmap_intercyl_hpf_{{tag}}.zarr"
```
and then add
```
- "ringmap_intercyl_hpf_lsd_{tag}.zarr"
```
to the `containers` block in the final `ZipZarrContainers` task.

As indicated in the comments, this will make a delay-filtered ringmap, and save 10 frequencies of that map to disk. The 10 frequencies were chosen to be outside of the persistent RFI bands and sample the full CHIME band. (I've checked in a handful of test days that these frequencies indeed survive all the previous masking.) In my testing, I've found that it's safer to turn off truncation for the map dataset, since some frequencies have many low-amplitude pixels that are zeroed by the truncation.

One could also consider taking the ringmaps produced by the daily pipeline and directly delay-filtering those, using `DayenuDelayFilterMap`, but to do this on a single day, you need to use `single_mask: False`, and I found that this blows up the memory usage and causes a crash.

The goal of these tasks is to have a handful of delay-filtered maps that we can use for validation for the intensity mapping analysis: systematics that affect that analysis sometimes only show up in filtered maps, so it will hopefully be informative to have a few filtered frequencies available for each day.

This should add ~5 minutes to the walltime for each day, and the final 10-frequency map will take 200-250MB of disk space (compared to ~100GB for all daily products).

jrs65:
@sjforeman do you have any indication that the truncation is doing anything beyond truncating pure noise? I don't think turning it off is an issue, but it'd be good to understand what the issue is.

sjforeman:
> @sjforeman do you have any indication that the truncation is doing anything beyond truncating pure noise? I don't think turning it off is an issue, but it'd be good to understand what the issue is.

Here's an example where the effect of truncation is notable:
Screen Shot 2023-03-13 at 10 46 54 PM
(I de-striped these maps by computing the median over nonzero-weight pixels at each elevation and subtracting that median.) I'm not sure whether it's pure noise that is being truncated, but there is interesting structure in the untruncated map that seems like it would be useful to retain.

For reference, here's the unfiltered map:
Screen Shot 2023-03-13 at 10 56 55 PM

ch_pipeline/core/dataquery.py
```diff
@@ -125,6 +125,7 @@ class QueryDatabase(task.MPILoggedTask):
     return_intervals = config.Property(proptype=bool, default=False)
 
     node_spoof = config.Property(proptype=dict, default=_DEFAULT_NODE_SPOOF)
+    connection_attempts = config.Property(proptype=int, default=1)
```
sjforeman, line 128: Needs a docstring entry

ch_pipeline/core/dataquery.py
```diff
@@ -516,6 +524,7 @@ class QueryAcquisitions(task.MPILoggedTask):
     """
 
     node_spoof = config.Property(proptype=dict, default=_DEFAULT_NODE_SPOOF)
+    connection_attempts = config.Property(proptype=int, default=1)
```
sjforeman, line 527: Also needs a docstring entry

ch_pipeline/core/dataquery.py
```diff
@@ -757,7 +766,7 @@ def next(
         return cont
 
 
-def finder_from_spec(spec, node_spoof=None):
+def finder_from_spec(spec, node_spoof=None, ntries=1):
```
sjforeman, line 769: Could consider adding docstring entries for `node_spoof` and `ntries`, and also fix typo in docstring: `dspec` should be `spec`

ch_pipeline/core/dataquery.py
```diff
@@ -808,7 +817,7 @@ def finder_from_spec(spec, node_spoof=None):
     return fi
 
 
-def files_from_spec(spec, node_spoof=None):
+def files_from_spec(spec, node_spoof=None, ntries=1):
```
sjforeman, line 820: Could consider adding docstring entries for `node_spoof` and `ntries`, and also fix typo in docstring: `dspec` should be `spec`

sjforeman (commented):
Just added a bunch of comments about docstrings that we might as well fix.

The daily pipeline config looks good to me. There has been some debate about the usefulness of outputting the delay-filtered ringmap at a handful of frequencies, and Arnab was assigned to look into alternatives. If he finds something better, we can change to that, but otherwise, I vote that we keep the delay-filtered ringmap outputs

PR #177, opened 2023-03-09
HFB use catalog

rikvl:


rikvl:
Relies on [ch_util PR #54](https://github.com/chime-experiment/ch_util/pull/54), so that would need to be looked at first.

This PR lets us provide the name of a source when loading files, which will automatically look up the source's declination (for selecting the NS beam in which it transits) and the frequency of its absorption feature (for frequency selection). Both these automatic selections can be overridden in various ways.

jrs65 (approved):

PR #178, opened 2023-03-13
fix(telescope): include correct complex check for `HEALPixBeam`s

sareda26:
The `beam` dataset in `HEALPixBeam` is a structured `np.dtype` that can't pass the `np.issubclass_` check. This patch extracts and checks the types of the `dtype` fields explicitly.


sjforeman (approved):
Looks good to me, thanks Alex!

PR #184, opened 2023-03-29
HFB general averaging

rikvl:
Replaces `HFBAverageOverTime` and `HFBAverageOverBeams` (which were very similar) with a general `HFBAverage` task that takes an `axis` attribute to specify over which axis to take the average. It can be expanded if we want to average over other axes (once the necessary containers are made).


rikvl:
I thought it was unnecessary to keep `average_hfb` as a separate function, so I moved it into the `HFBAverage` task.

This is ready for review.

jrs65 (approved):

PR #185, opened 2023-04-04
feat(flagging): add a task to expand gain masks in time

ljgray:
This PR modifies the implementation of the `flagging.ExpandMask` task to allow custom behaviour when creating the mask to expand. It also adds a task which will expand a gain mask only in the places where the gains are flagged due to bad calibration, which was the intended behaviour of the gain masking task.

This works as expected but I'm not convinced that it's actually doing much that the rfi flagging wouldn't catch anyway

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -1935,11 +1935,12 @@ class ExpandMask(task.SingleTask):
         -------
         exp_mask : RFIMask or SiderealRFIMask
         """
-
-        nfreq, ntime = raw_mask.mask[:].shape
-
-        mraw = np.zeros((nfreq, ntime + 2 * self.nexpand), dtype=bool)
-        mraw[:, self.nexpand : -self.nexpand] = raw_mask.mask[:]
+        # Get a reference to the mask
+        mask = raw_mask.mask[:]
+        # Log the original fraction of flagged data
+        original_mask_sum = mask.sum()
+        # Produce the padded mask to expand in time
+        mraw = self._padded_mask_hook(mask)
 
         window = 2 * self.nexpand + 1
         mexp = np.any(rfi._rolling_window_lastaxis(mraw, window), axis=-1)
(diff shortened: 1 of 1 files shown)


ch_pipeline/analysis/flagging.py
```diff
@@ -1948,11 +1949,57 @@ def process(self, raw_mask):
...
+    def _padded_mask_hook(self, mask: np.ndarray) -> np.ndarray:
+        """Create the padded mask to expand."""
+        # Initialize the empty mask. We don't want to combine with the
+        # original mask until after expanding
+        nfreq, ntime = mask.shape
+        mraw = np.zeros((nfreq, ntime + 2 * self.nexpand), dtype=bool)
+
+        # Indices of the first and last unmasked samples along the time axis.
+        time_start_idx = mask.argmin(axis=1) + self.nexpand
+        time_end_idx = mraw.shape[1] - mask[:, ::-1].argmin(axis=1) - self.nexpand
```
sjforeman, line 1991: In cases where every sample at a given frequency is masked, these will spuriously indicate that the first and last samples are unmasked

ch_pipeline/analysis/flagging.py
```diff
@@ -1948,11 +1949,57 @@ def process(self, raw_mask):
...
+
+        # Only take frequencies where there is masking in the first or last samples
+        freq_start_idx = time_start_idx > self.nexpand
+        freq_end_idx = time_end_idx < mraw.shape[1] - self.nexpand
```
sjforeman, line 1995: I find these variable names confusing, since `freq_start_idx` and `freq_end_idx` are actually boolean masks, while `time_start_idx` and `time_end_idx` are arrays of indices.

sjforeman (changes requested):
Looks good overall - just a few small comments

ljgray:
I'm actually just going to scrap this PR for now. The logic isn't great and it doesn't really align with the intended goal of only masking around bad gain calibrations

PR #187, opened 2023-04-25
fix(dataquery): handle case where no fmap is found for a given dataset_id

ljgray:
Two quick fixes:
- dataset id has no associated frequency map (use the default frequency map)
- dataset id not found in chime database (if none of the available dataset ids are found, raise an exception)

### DIFF
--- a/ch_pipeline/core/dataquery.py
+++ b/ch_pipeline/core/dataquery.py
@@ -690,6 +690,7 @@ class QueryFrequencyMap(task.MPILoggedTask):
         freq_map
             Frequency slot map container
         """
+        from peewee import DoesNotExist
 
         # Fetch from the cache if we can
         if self.cache and self._cached_inputs:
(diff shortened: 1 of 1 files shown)


ch_pipeline/core/dataquery.py
```diff
@@ -711,25 +713,43 @@ def next(
...
+                    except exceptions.NotFoundError:
+                        # No frequency map found for this dataset id, so the default
+                        # is used. Provide an entry so we can still check how many
+                        # frequency maps were found
+                        fmaps["default"] = 1
+                    except DoesNotExist:
+                        # This dataset id does not exist in the database. If none of the
+                        # dataset ids are found, throw an exception
+                        _missing_ds_flag = True
+                        continue
```
jrs65, line 736: Peewee could throw a `DoesNotExist` at many stages. I think being a little more surgical with the exception checking might help here. Something like:

```suggestion
                    _missing_ds_flag = False
                    try:
                        dataset = ds.Dataset.from_id(str(dsid))
                    except DoesNotExist:
                        # This dataset id does not exist in the database. If none of the
                        # dataset ids are found, throw an exception
                        _missing_ds_flag = True
                        continue
                        
                    try:
                        state = dataset.closest_ancestor_of_type("f_engine_frequency_map").state
                        fmaps[state.id] = state.data
                    except exceptions.NotFoundError:
                        # No frequency map found for this dataset id, so the default
                        # is used. Provide an entry so we can still check how many
                        # frequency maps were found
                        fmaps["default"] = 1
```

jrs65 (dismissed):
I left an optional suggest, but this looks pretty good so just merge when you're happy.

ch_pipeline/core/dataquery.py
ljgray: Good idea

PR #189, opened 2023-04-27
feat(daily): save beamformed visibilities versus hour angle

ssiegelx:
Save the beamformed visibilities versus hour angle for the bright source catalog (ps_cora_10Jy) to be used for offline stability analysis.

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -423,6 +423,23 @@ pipeline:
         save: true
         output_name: "sourceflux_{{tag}}.h5"
 
+    # Load the source catalogs to measure flux as a function of hour angle
+    - type: draco.core.io.LoadBasicCont
+      out: source_catalog_nocollapse
+      params:
+        files:
+        - "{catalogs[0]}"
+
+    # Measure the beamformed visibility as a function of hour angle
+    - type: draco.analysis.beamform.BeamFormCat
+      requires: [manager, sstream_inter]
+      in: source_catalog_nocollapse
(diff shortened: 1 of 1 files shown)


ssiegelx:
@ljgray I'm not sure if it is too late to have this merged for rev 07?

I would like to save the beamformed visibilites for the catalog of bright sources as a function of hour angle in addition to collapsing over the hour angle axis.

I would like to use the resulting data products to identify and mask frequencies that are suffering from narrowband multiplicative errors.  If we collapse over hour angle it is difficult to separate these types of features from transient RFI -- either unmasked RFI that is biasing the average or masked RFI coupling hour angle dependent errors in the beam model.

If we save 300 seconds on either side of transit that should yield a data product that is ~150 MB per day.

I've kept the version that is collaped over hour angle because it is cheap and I imagine others might prefer to work with this simpler version.

ljgray (approved):
No issues including this in rev_07, I'll merge this in when I merge the rest of the config

ssiegelx:
Awesome, thank you!

ljgray:
I actually had to make a tweak to this config to avoid having the second `BeamFormCat` task run setup before the first one finished

ljgray:
@jrs65 @ssiegelx I had to add the call to `WaitUntil` and run the non-collapsed task first but it should work now. The issue was both `BeamFormCat` tasks calling setup before either had processed, meaning that each one a copy of the visibility dataset stored at the same time

jrs65 (approved):

PR #191, opened 2023-04-28
feat(daily): add reduced ringmaps to daily config

ljgray:
Goes along with radiocosmology/draco#241


sjforeman (changes requested):
Could we save these variances for both the XX and YY pols? We've seen interesting differences between XX and YY in previous revisions, so having the diagnostic for both pols might be useful

PR #192, opened 2023-04-28
feat(daily): add reduced ringmaps to daily config

ljgray:
Somehow the previous PR wouldn't reflect updates to the branch, so it's easier to just make a new one since this is minor.

Goes with radiocosmology/draco#241

Updated to include both XX and YY pol

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -649,7 +649,51 @@ pipeline:
         output_name: "ringmap_intercyl_hpf_{{tag}}.zarr.zip"
         remove: true
 
+    # Downselect the ringmap to keep only the XX and YY pols
+    - type: draco.analysis.transform.Downselect
+      in: ringmap_int_hpf
+      out: ringmap_int_hpf_xx_yy
+      params:
+        selections:
+          pol_index: [0, 3]
+
+    # Take the variance of the map across elevation. Apply weights
+    # as a mask only
+    - type: draco.analysis.transform.ReduceVar
+      in: ringmap_int_hpf_xx_yy
+      params:
+        axes:
+          - el
+        dataset: map
+        weighting: weighted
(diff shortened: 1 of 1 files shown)


ch_pipeline/processing/daily.py
```diff
@@ -622,7 +622,75 @@
...
+    - type: draco.analysis.transform.Reduce
+      in: ringmap_int_hpf
+      out: ringmap_reduced_freq
+      params:
+        axes: ["freq"]
+        datasets: ["map"]
+        op: "var"
+        selections: {"pol":[0, 3]}
+        generate_weight: true
+        apply_weight: false
```
sjforeman, line 636: Making sure params agree with `Reduce` task:
```suggestion
        generate_mask: true
        apply_mask: false
```

ch_pipeline/processing/daily.py
```diff
@@ -622,7 +622,75 @@
...
+    - type: draco.analysis.transform.Reduce
+      in: ringmap_int_hpf
+      out: ringmap_reduced_el
+      params:
+          axes: ["el"]
+          datasets: ["map"]
+          op: "var"
+          selections: {"pol":[0, 3]}
+          generate_weight: true
+          apply_weight: false
```
sjforeman, line 668: Making sure params agree with `Reduce` task:
```suggestion
          generate_mask: true
          apply_mask: false
```

sjforeman (approved):
This looks good after fixing a few of the `Reduce` params

ch_pipeline/processing/daily.py
ljgray: Good catch, thanks. I might actually have to change this again after reworking the `Reduce` task a bit

sjforeman (approved):
Looks good to merge after the corresponding draco PR

PR #196, opened 2023-05-24
Pull HFB staging into master

jrs65:


rikvl (approved):
I resolved the merge conflicts. Apart from one, they seemed more like hiccups where I thought they should have merged automatically. Either way, I think this is now ready to be merged.

rikvl:
I thought this should be a "rebase and merge" to keep some of the HFB commit history intact, but now GitHub tells me "This branch cannot be rebased due to conflicts". @jrs65, did you want to "squash and merge" or should we rebase `hfb-staging` onto `master`?

jrs65:
I don't think we want to squash and merge as we'll lose all the history. I think the best thing to do is for me to force a normal merge (ignoring the linear history requirement).

rikvl:
Ah, if that's an option, that would be preferable in this case. For me GitHub says that "create a merge commit" is "not enabled for this branch". Do you have privileges to force it? If so, go ahead!

PR #197, opened 2023-06-01
feat(hfb): add container_time to attrs

rikvl:
This is usually the transit time, useful to have quick access to.


rikvl:
Don't do this, because the time can be calculated from the LSD and source, and it's generally bad to have duplicate information

PR #202, opened 2023-06-22
refactor(hfb): create base HFB container

rikvl:
PR #200 should be handled before this one.

I gathered the properties that were repeated for all HFB containers (except the `HFBRFIMask` one) into a new base HFB container class called `HFBContainer`. There was some inconsistency in the naming of the weights dataset between different containers. I tried to handle this, but wasn't sure if the existing names had a purpose or if some functionality could break because of my changes.

All HFB containers so far were subclasses of `core.containers.FreqContainer`, so I made `HFBContainer` a subclass of `FreqContainer`. However, HFB data does not necessarily need to have a frequency axis (although its primary purpose requires keeping a frequency axis). Perhaps at some point we want to make a container for light curves, integrated over frequency, or something like that. In that case, it would make more sense to make `HFBContainer` a subclass of `core.containers.ContainerBase`, and keep the current HFB container classes explicitly a subclass of `FreqContainer` (as well as a subclass of the new `HFBContainer`). What do you think?

### DIFF
--- a/ch_pipeline/hfb/containers.py
+++ b/ch_pipeline/hfb/containers.py
@@ -245,8 +245,12 @@ class HFBTimeAverage(FreqContainer, HFBContainer):
     }
 
 
-class HFBHighResData(TODContainer, FreqContainer, HFBContainer):
-    """Container for holding high-resolution frequency data"""
+class HFBHighResContainer(FreqContainer, HFBContainer):
+    """Base class for HFB containers with high-resolution frequency data."""
+
+
+class HFBHighResData(TODContainer, HFBHighResContainer):
+    """Container for holding high-resolution frequency data."""
 
     _axes = ("beam",)
 
@@ -274,8 +278,8 @@ class HFBHighResData(TODContainer, FreqContainer, HFBContainer):
     }
 
 
-class HFBHighResTimeAverage(FreqContainer, HFBContainer):
-    """Container for holding time-averaged high-resolution frequency data"""
+class HFBHighResTimeAverage(HFBHighResContainer):
+    """Container for holding time-averaged high-resolution frequency data."""
 
     _axes = ("beam",)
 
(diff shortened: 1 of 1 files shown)


ch_pipeline/hfb/containers.py
```diff
@@ -10,12 +10,40 @@
...
+
+    @property
+    def weight(self) -> memh5.MemDataset:
+        """The inverse variance weight dataset."""
+        if "weight" in self:
+            weight = self["weight"]
+        elif "hfb_weight" in self:
+            weight = self["hfb_weight"]
+        else:
+            weight = self["flags/hfb_weight"]
```
jrs65, line 37: This access will only work on subclasses of RawContainer and I believe will throw an exception as access to other groups is not allowed without `RawContainer`. You might want to protect the call a bit more.

ch_pipeline/hfb/containers.py
```diff
@@ -10,12 +10,40 @@
...
+class HFBContainer(ContainerBase):
+    """A base class for all HFB containers.
+
+    Like :class:`ContainerBase`, but with some properties specific to HFB data.
+    """
+
+    @property
+    def hfb(self) -> memh5.MemDataset:
+        """The main hfb dataset."""
+        return self.datasets["hfb"]
```
jrs65, line 27: Given that you're centralising the logic here, you might want to actually improve on it and offer a more useful exception message if this lookup fails.

jrs65 (approved):
Nice, a good clean up. A few optional suggestions in comments.

ch_pipeline/hfb/containers.py
rikvl: Done.

ch_pipeline/hfb/containers.py
rikvl: Done.

jrs65 (approved):

PR #204, opened 2023-07-04
fix(daily): add longer buffer to recent day list

ljgray:
Set a buffer two days back before processing a CSD to ensure that all data has been transferred to cedar. This should probably be temporary until a better way to check for available data is in place

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -984,11 +984,13 @@ class DailyProcessing(base.ProcessingType):
         to_run = self.status(user=user)["not_yet_submitted"]
 
         if self._num_recent_days:
-            today = math.floor(ephemeris.chime.get_current_lsd())
+            buffer = 2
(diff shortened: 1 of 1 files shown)


jrs65:
What's the lag time for a file to be registered?

I thought what it's meant to do is find all the files registered in the db that would span that day, and then only run if all of them are on cedar. It usually only takes an hour or so for a file to be registered so I'm surprised that logic wouldn't work.

ljgray:
> What's the lag time for a file to be registered?
> 
> I thought what it's meant to do is find all the files registered in the db that would span that day, and then only run if all of them are on cedar. It usually only takes an hour or so for a file to be registered so I'm surprised that logic wouldn't work.

This is actually because the weather data seems to be taking longer to transfer over, which has been causing all recent days to fail until I re-run them the next day. I'm probably going to expand the file checking for available days to also include other relevant files like the weather, so this is just a hotfix for the time being

jrs65:
Ah okay. Sorry. So this is the weather data issue which makes sense.

PR #205, opened 2023-07-31
feat(hfb): add HFB RingMap containers

rikvl:


ch_pipeline/hfb/containers.py
```diff
@@ -338,3 +338,72 @@ class HFBHighResSpectrum(HFBHighResContainer):
...
+            "dtype": np.uint16,
+            "initialise": False,
+            "distributed": True,
+            "distributed_axis": "freq",
+        },
+    }
+
+
+class HFBHighResRingMap(FreqContainer, HFBRingMapBase, HFBHighResContainer):
+    """Container for holding high-resolution frequency ringmap data."""
```
jrs65, line 385: I would maybe add a note in the documentation to emphasise that freq has moved to the back, and give some information as to why that's not insane.

jrs65 (approved):
Good, with one suggested comment.

PR #206, opened 2023-07-31
feat(hfb): stitch freqs together for HFB ringmaps

rikvl:


ch_pipeline/hfb/analysis.py
```diff
@@ -195,6 +195,70 @@ def process(self, stream):
...
+        nsubfreq = len(subfreq)
+        new_shape = (nbeam, None, nra, nfreq * nsubfreq)
+
+        # Combine frequency and sub-frequency axes
+        data = data.reshape(new_shape)
+        weight = weight.reshape(new_shape)
+
+        # Create container to hold output
+        out = containers.HFBHighResRingMap(
+            beam=beam, el=el, ra=ra, freq=freq, attrs_from=stream
```
jrs65, line 251: Why not use `axes_from`? That will just copy over the el and ra axes.

jrs65 (changes requested):

ch_pipeline/hfb/analysis.py
rikvl: Right, that fix ended up in a different branch. I added it here now.

jrs65 (approved):

arashcosmology (approved):

PR #209, opened 2023-08-03
fix(hfb.io): allow loading of a simple list of files

jrs65:


rikvl (approved):
Checked this -- looks good!

PR #210, opened 2023-08-22
HFB regridding

jrs65:
- feat(hfb.containers): use DataWeightContainer as base for generic functionality
- feat(HFBData): allow distributed load with from file. TEMPORARY
- feat(HFBSiderealRegridder): specialised HFB regridder

### DIFF
--- a/ch_pipeline/hfb/sidereal.py
+++ b/ch_pipeline/hfb/sidereal.py
@@ -5,11 +5,11 @@ from caput import mpiarray
 from ch_util import ephemeris
 from draco.analysis.sidereal import SiderealRegridderLinear
 
+from beam_model.formed import FFTFormedActualBeamModel
+
 from .containers import HFBData, HFBRingMap
 
 
-# TODO: this is really just a copy-paste of much of the underlying .process(...), with
-# some refactoring of the base task this task could probably be eliminated
 class HFBSiderealRegridder(SiderealRegridderLinear):
     """Regrid HFB data."""
 
@@ -24,6 +24,10 @@ class HFBSiderealRegridder(SiderealRegridderLinear):
 
         # Set up the default Observer
         self.observer = ephemeris.chime if observer is None else observer
+
+        # Get hour angles of EW beams (in deg) from beam model
+        beam_mdl = FFTFormedActualBeamModel()
+        self.ew_beam_offset_deg = beam_mdl.config["ew_spacing"]
 
     def process(self, data: HFBData) -> HFBRingMap:
         """Regrid HFB timestream data onto the sidereal day.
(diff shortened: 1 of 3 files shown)


rikvl:
I was looking into creating the `el` axis (one of the `TODO`s in this PR). I believe something like this could do that:
```
import numpy as np

from beam_model.formed import FFTFormedActualBeamModel

ns_beams = np.array([105, 118, 127])
freq = 600.0

mdl = FFTFormedActualBeamModel()
za_deg = mdl.get_beam_positions(ns_beams, freq)[..., 1].squeeze()

el = np.sin(za_deg * np.pi / 180.0)
```
However, this reveals that the NS pointing of the formed beam depends on frequency.

Would it be better to abandon the `el` axis, replacing the `beam` and `el` axes in `HFBRingMap` with `ew_beam` and `ns_beam` axes, and then create some helper functions (with a frequency argument) to convert NS beam indices to el or dec? Or should we fudge the el? Or am I missing something?

rikvl:
The variation in NS pointing with frequency seems to be of the order of 0.1 arcmin, similar to the spacing of the beams.

Here's the NS pointing as function of frequency for NS beams 120 through 127
za-freq-ns_beam120-127

rikvl:
There was something wrong with the figure in the previous comment.

Here's an updated version of the frequency dependence of the zenith angle of synthetic beams, also showing the reference angles. This indicates that, as expected, the reference angles correspond to the ZA around which the pointing varies.

za-freq-ns_beam120-127-ref

rikvl:
@jrs65 I implemented the following:
- An extra axis was added to the HFB ring map containers (in `HFBRingMapBase`) to store the NS beam index (because the `el` axis does not reflect the true el pointing of the beam for most frequencies),
- The different EW beams are regridded separately, with each its own offsets in time derived from its offset in the CHIME/FRB beam-model x coordinate: (-0.4, 0, 0.4, 0.8) deg for EW beam 0 through 3, respectively.

This way of dealing with the off-meridian beams works only approximately for most beams, and breaks down completely near the NCP, I believe. The time offset should depend on both EW and NS index. But even then there the issue that closer to the NCP, a given NS beam index corresponds to different Declinations for the different EW beams, while in the current scheme we assign them to the same el, with the idea of combining or comparing the different EW beams

I think we need to discuss whether it makes sense to use an el = sin(ZA) axis when dealing with off-meridian EW beams. The following plot may help:
beams-isodec-01

rikvl:
Out of all the recent HFB PRs, I think this one has the highest priority, because some other PRs depend on it.

@jrs65, I added code to this PR and wanted to request a review from you, but that didn't seem possible because you originally started the PR. Can you somehow start a review from yourself, or otherwise provide feedback outside of a Github review?

ch_pipeline/hfb/sidereal.py
```diff
@@ -0,0 +1,114 @@
+"""Regrid the data to sidereal time."""
+import numpy as np
```
rikvl, line 2:
```suggestion
"""Regrid the data to sidereal time."""

import numpy as np
```

New version of `black` wants a blank line here.

rikvl (approved):

rikvl:
Please hold off merging this for now. It looks like some commits got lost that are needed in the regridding. I'll fix this tonight or tomorrow.

jrs65:
Okay. That's weird. I did have to remove some commits from the rebase because they were already in master (seemingly), but I went through fairly thoroughly to make sure all the relevant changes were present. Sorry if I missed something.

jrs65:
This looks good to me. Rebase it on the latest master, and then feel free to do a rebase and merge.

PR #211, opened 2023-08-28
Wildcard file extension to pick up both .h5 and .zarr.zip files

ljgray:


jrs65:
@ljgray you should work off the branch `quarterstack` that has a bunch of fixes (including one for this).

ljgray:
> @ljgray you should work off the branch `quarterstack` that has a bunch of fixes (including one for this).

Ah, great

PR #214, opened 2023-10-05
Allow selections when loading CHIME data

jrs65:
- feat(MakeTelescope): a task to construct a telescope object directly
- feat(LoadCorrDataFiles): allow general axis selections

### DIFF
--- a/ch_pipeline/core/telescope.py
+++ b/ch_pipeline/core/telescope.py
@@ -13,13 +13,14 @@ import h5py
 import healpy
 from scipy.interpolate import RectBivariateSpline
 
-from caput import config, mpiutil
+from caput import config, misc, mpiutil
 
 from cora.util import coord, hputil
 
 from drift.core import telescope
 from drift.telescope import cylbeam
 
+from draco.core import task
 from draco.core.containers import ContainerBase, GridBeam, HEALPixBeam
 
 from ch_util import ephemeris, tools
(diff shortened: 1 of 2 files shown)


ljgray (approved):

ch_pipeline/core/telescope.py
```diff
@@ -1040,3 +1041,40 @@ def _nearest_freq(tel_freq, map_freq, freq_id, single=False):
...
+    telescope_type : str
+        The type of telescope object to create. Generally this must be a fully qualified
+        Python class name, however, \"chime\" (default) can be used to create a
+        `ch_pipeline.core.telescope.CHIME` instance.
+    telescope_config : dict
+        Configuration passed straight to the telescope class via it's `from_config`
+        method.
+    """
+
+    telescope_type = config.Property(proptype=str, default="CHIME")
```
ljgray, line 1063:
```suggestion
    telescope_type = config.Property(proptype=str, default="chime")
```

ljgray (changes requested):

ch_pipeline/core/telescope.py
```diff
@@ -1040,3 +1041,40 @@ def _nearest_freq(tel_freq, map_freq, freq_id, single=False):
...
+
+    def setup(self) -> telescope.TransitTelescope:
+        """Create and return the telescope object."""
+
+        _type_map = {
+            "chime": CHIME,
+        }
+
+        if self.telescope_type in _type_map:
+            tel_class = _type_map[self.telescope_type.lower()]
```
ljgray, line 1074: This seems a bit redundant - if `telescope_type` and `telescope_type.lower()` aren't the same then this shouldn't work. A "try - except" block or "_type_map.get" would make more sense I think

ljgray (changes requested):

ch_pipeline/core/telescope.py
jrs65: I've just removed the lower casing. Given that this task didn't exist before it's not like there's any legacy behaviour to account for. People should just use the right names.

PR #215, opened 2023-10-20
Include module load options and defaults

ljgray:
Requires radiocosmology/caput#255

### DIFF
--- a/ch_pipeline/processing/base.py
+++ b/ch_pipeline/processing/base.py
@@ -16,6 +16,8 @@ cluster:
   temp_directory: {tempdir}
 
   venv: {venv}
+  module_path: {modpath}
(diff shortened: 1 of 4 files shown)


rikvl (changes requested):
I believe the path to the module files is:
`/project/rpp-chime/chime/chime_env/modules/modulefiles`
Sorry, this may have changed in the meantime because it took so long for me to review the PR

ljgray:
> I believe the path to the module files is: `/project/rpp-chime/chime/chime_env/modules/modulefiles` Sorry, this may have changed in the meantime because it took so long for me to review the PR

Nope, you're right. I'm not entirely sure how that happened - I tested this code and must have screwed it up in a rebase or something

PR #219, opened 2023-11-21
fix(hfb.analysis): fix shape of pfb_shape

rikvl:
This will make it work for the blind-search pipeline, but it breaks the targeted pipeline. It shouldn't be hard to make it work for both pipeline versions.

Making this draft PR so we don't forget about this.


rikvl:
I made a change to make `HFBDividePFB` work for both `HFBTimeAverage` containers (needed for the targeted pipeline) and `HFBRingMap` containers (needed for the blind-search pipeline). This is currently implemented as an if statement checking the container type. If you think it's better to implement it differently, let me know.

jrs65 (approved):

arashcosmology (approved):

PR #222, opened 2023-11-23
HFB add beam container

rikvl:
Make a base class `HFBBeamContainer` for all HFB containers that have
that have a `beam` axis. This also has `beam_ew` and `beam_ns`

I think this code could be looked at already, ~~but merging may need to wait for PR #210 to be merged~~

Note to self: ~~after PR #210 is handled, but~~ before merging this PR, check if any changes still need to be added in `hfb/analysis.py` from commit 21a267c8 that could otherwise get lost

PR #224, opened 2023-11-23
HFB averaging high-res ringmaps over EW beams

rikvl:
This code depends on PR #222 (HFB beam container), which seems to depend on PR #210 (regridding).

Also, the averaging task is not of much use without PR #223 (aligning EW beams)

### DIFF
--- a/ch_pipeline/hfb/containers.py
+++ b/ch_pipeline/hfb/containers.py
@@ -365,7 +365,7 @@ class HFBHighResSpectrum(HFBHighResContainer):
 class HFBRingMapBase(SiderealContainer, HFBContainer):
     """Base class for HFB ringmaps.
 
-    These containers include axes to mark the EW and NS beam indices, as well as
+    This container includes an axis to mark the indices of the NS beams, as well as
     RA (inherited from :class:`SiderealContainer`) and el = sin(zenith angle) axes.
 
     The el axis corresponds to the sin(za) of the reference angles for the NS beams.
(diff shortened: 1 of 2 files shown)


ch_pipeline/hfb/containers.py
```diff
@@ -51,7 +54,50 @@ def nsample(self) -> memh5.MemDataset:
...
+        return np.unique(self.beam // 256, return_inverse=return_inverse)
+
+    def get_beam_ns(self, return_inverse=False):
+        """Find the unique NS-beam indices (i.e., from 0 to 256) in the beam axis.
+
+        If called with `return_inverse=True`, return a tuple of (a) unique NS-beam
+        indices and (b) an array mapping each sample in the beam axis to an index
+        in the output array of unique NS-beam indices.
+        """
+        return np.unique(self.beam % 256, return_inverse=return_inverse)
```
jrs65, line 97: I suggest having only the properties and remove the get methods to tighten up the API. Also, you should cache the properties to avoid regenerating at every call. There is a `cached_property` decorator for that in `functools` that we use elsewhere in the pipeline.

jrs65 (changes requested):
Is this PR pointed at the right base branch? There's a bunch of changes which look like they belong in other PRs?

rikvl:
> Is this PR pointed at the right base branch? There's a bunch of changes which look like they belong in other PRs?

Changed the base branch to `rvl/hfb-beam-container`. Sorry about that, I think you already explained this on Tuesday.

ch_pipeline/hfb/containers.py
rikvl: These changes are now implemented in PR #222

jrs65 (approved):

arashcosmology (approved):

PR #225, opened 2023-11-23
HFB weighted median subtraction 

arashcosmology:


ch_pipeline/hfb/analysis.py
```diff
@@ -936,3 +937,59 @@ def _ensure_list(x):
...
+        # Change the order of axes in data and mask arrays, as weighted median
+        # is calculated along the last axis
+        data_s = np.swapaxes(data, 2, 3)
+        binary_weight = np.swapaxes(binary_weight, 2, 3)
+
+        # Calculate weighted median (to exclude flagged data) along beam axis:
+        median = weighted_median.weighted_median(data_s, binary_weight)
+
+        # Subtract weighted median along all beams from the data
+        diff = data - median[:, :, None, :]
```
rikvl, line 986:
```suggestion
        diff = data - median[:, :, np.newaxis, :]
```

ch_pipeline/hfb/analysis.py
```diff
@@ -936,3 +937,59 @@ def _ensure_list(x):
...
+
+        # Create container to hold output
+        out = containers.HFBData(stream)
+
+        # Place diff in output container
+        out.hfb[:] = diff[:]
+        out.weight[:] = weight[:]
```
rikvl, line 993:
```suggestion
        out.hfb[:] = diff
        out.weight[:] = weight
```
Not 100% sure, but I think the `[:]` is not necessary

rikvl (changes requested):

ch_pipeline/hfb/analysis.py
```diff
@@ -1047,3 +1048,59 @@ def _ensure_list(x):
...
+        # Change data to numpy array, so that it can be reshaped
+        if isinstance(data, mpiarray.MPIArray):
+            data = data.local_array
+            weight = weight.local_array
+
+        # make a mask of non-zero weights
+        mask = weight != 0
+
+        # Generate binary weight
+        binary_weight = mask.astype(float)
```
jrs65, line 1086: I prefer to explicitly set the precision. In this case `np.float32` would suffice.

ch_pipeline/hfb/analysis.py
```diff
@@ -1047,3 +1048,59 @@ def _ensure_list(x):
...
+
+
+class HFBMedianSubtraction(task.SingleTask):
+    """Subtracting weighted median along beam axis to remove fluctuations in the
+    data induced by temperature fluctuations in East and West receiver huts."""
+
+    def process(self, stream):
+        """Subtract weighted median of the data along beam axis from the data.
+        A binary mask (0 when the weight is zero, and 1 for non-zero weights)
+        is used in weighted median
```
jrs65, line 1060: There is a convention for docstrings that should be followed. They should be a one line short description (max 88 chars), and then additional details in a subsequent paragraph.

jrs65 (changes requested):

rikvl (approved):

PR #230, opened 2024-01-16
feat(hfb): add Doppler shift for HFB ringmap

rikvl:
Currently looping over `ra` and `el` axes.


rikvl:
Maybe it's better to avoid looping over `ra` and `el`. By evaluating at a single time this can be avoided. Doppler shift due to Earth's motion may vary slow enough that we only need one evaluation per day/week. See if performance requires this.

Issue #237, opened 2024-02-16
HFB pipeline hangs while writing file if running on multiple nodes


rikvl:
I did a bunch more tests of running the HFB blind-search pipeline on 2 nodes to try to debug the MPI issue that manifests as the pipeline getting stuck on writing files, specifically in the task `MakeHighFreqResRingMap` (where the distributed axis changes from `freq` to `el`, i.e., NS beam).

It seems the problem goes away when making sure that (1) the number of frequency channels and (2) the number of NS beams in the data are both an integer multiple of the number of MPI processes (i.e., `nodes` x `pernode`).

However, when I just did one last big test to confirm that the pipeline runs under these circumstances, it suddenly got stuck while writing in the SiderealGrouper task, but only in the 4th and final sidereal day (LSD 2895) being processed. Hopefully that was a glitch or a different problem.

Attached are two pipeline config files to reproduce the issue (changed yaml to txt extension to add them to this GH issue), a minimal one that loads files produced by the task just before `MakeHighFreqResRingMap` and a full one that runs through the entire pipeline.
[debug-mpi-minimal.txt](https://github.com/chime-experiment/ch_pipeline/files/14307361/debug-mpi-minimal.txt)
[debug-mpi-full.txt](https://github.com/chime-experiment/ch_pipeline/files/14307360/debug-mpi-full.txt)

Branches / commits I was using:
```
caput                                    master            bea4d36a69d7befe4cac2a84e693ac7d3be5db74
ch_pipeline                              rvl/hfb-mpi-test  761223051d40597d5d86dc0baf749e2fbc127023
ch_util                                  rvl/hfb-testing   7d25e2032a5cf627a4275fa3e42ae3bd4b616c29
draco                                    rvl/hfb-test      ab06bcf62c798fb5a7288739de9fff41df567c06
```

From `draco.core.misc.DebugInfo`:
```
caput                                    version=23.6.0+14.g364df19
ch_pipeline                              version=0+untagged.627.g247465e
ch_util                                  version=23.6.0+12.g2db1fb0
chimedb                                  version=23.8.0
chimedb.config                           version=21.1.0
chimedb.data-index                       version=23.6.0+2.ga64cf19
chimedb.dataflag                         version=23.6.0+4.gf962b54
chimedb.dataset                          version=23.10.0
cora                                     version=22.9.0
draco                                    version=23.6.0+53.g9df3973
driftscan                                version=22.9.0
```

jrs65:
[...]

PR #238, opened 2024-02-27
HFB MPI fix hang on write-out

rikvl:
Workaround for #237


rikvl:
To summarize what we learned last February (see comments in #237): running the HFB blind-search pipeline on multiple nodes hangs in a file write-out stage when the number of MPI processes is larger than the number of NS beams, possibly due to some upstream bug. This PR is a work-around to avoid the failure conditions.

ljgray (approved):
This workaround looks fine. I think that this issue is probably the same as the one discussed in radiocosmology/caput#165, which had a lot of investigation but was never resolved.

PR #240, opened 2024-05-02
Update quarterstack config for daily pipeline rev_08 changes

ljgray:


ch_pipeline/processing/quarterstack.py
```diff
@@ -219,14 +198,34 @@
...
+      params:
+        source: ["CAS_A", "CYG_A", "TAU_A", "VIR_A"]
+
+    # Try and derive an optimal time-freq factorizable mask that covers the
+    # existing masked entries
+    - type: draco.analysis.flagging.MaskFreq
+      in: sstack_flag_src
+      out: factmask
+      params:
+        factorize: true
```
Arnab-half-blood-prince, line 218: don't we want to mention the known freq band which are bad. 

bad_freq_ind: [[738, 753], [703, 720], [490, 509], [857, 873], [485, 488]]

ljgray: All of these bands should now be in the static mask, so I think as long as we aren't doing any blending then we don't need to mask them. 

I'm also thinking that for the stacks it's better not to do any additional masking that isn't in the actual quarterstack files, so that we can easily see if there's a bad a band that made it through before stacking

sjforeman:
Suggested changes for updated rainfall flagging:
1. Remove `rain1mm_sd` from flags in `type: ch_pipeline.analysis.flagging.DataFlagger` block
2. Add `FlagRainfall` (https://github.com/chime-experiment/ch_pipeline/pull/244) after `DataFlagger`:
```
    - type: ch_pipeline.analysis.flagging.FlagRainfall
      in: sstream_mask3
      out: sstream_mask4
      params:
        accumulation_time: 30.0
        threshold: 1.0
```
3. Updated later `sstream_maskN` references for consistency

ljgray:
@sjforeman Although we aren't using this much at the moment, I'd like to get these changes merged in so that we stay reasonably up to date

PR #241, opened 2024-05-03
Chime-specific version of new RFI mask

ljgray:
Based on radiocosmology/draco#265. Requires 265 to merge for tests to pass

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -210,6 +210,47 @@ class RFISensitivityMask(dflagging.RFISensitivityMask):
         return ~rfi.frequency_mask(freq, timestamp=timestamp)
 
 
+class RFIStaticMask(task.SingleTask):
+    """Get the static mask for the time period covered by the data.
+
+    This is the same static mask used in :class:`RFIFilter` and
+    :class:`RFISensitivityMask`.
+    """
+
+    def process(self, data):
+        """Create a mask with all static frequency bands for a given day.
+
+        Parameters
+        ----------
+        data
+            container with data to mask. Should have either a time-like axis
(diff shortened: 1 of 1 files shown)


ch_pipeline/analysis/flagging.py
```diff
@@ -210,6 +210,44 @@ def _static_rfi_mask_hook(self, freq, timestamp=None):
         return ~rfi.frequency_mask(freq, timestamp=timestamp)
 
 
+class MaskStaticRFI(task.SingleTask):
```
sjforeman, line 272: Should this be called `StaticRFIMask` or something similar, since it generates a mask but doesn't actually apply the mask?

sjforeman: Or maybe `GetStaticRFIMask`...

ch_pipeline/analysis/flagging.py
```diff
@@ -141,6 +142,64 @@ def process(self, data) -> Union[containers.RFIMask, dcontainers.RFIMask]:
...
+
+        Returns
+        -------
+        mask : np.ndarray[nfreq]
+            Mask array. True will mask a frequency channel.
+        """
+        return rfi.frequency_mask(freq, timestamp=timestamp)
+
+    def _source_flag_hook(self, times):
+        """Flag times where bright sources are transiting.
```
sjforeman, line 173:
```suggestion
        """Flag times where bright sources are transiting or sun is up.
```

sjforeman (approved):
Only minor comments

ch_pipeline/analysis/flagging.py
ljgray: I guess the current standard is `RFI___Mask` so I've renamed it to follow that

PR #242, opened 2024-05-07
feat(flagging): apply masks to rebin_weight datasets

ljgray:


ljgray:
I think this should catch everything in the quarter stack/full stack

ch_pipeline/analysis/flagging.py
```diff
@@ -2121,6 +2124,9 @@ def process(self, timestream):
             ]
             tools.apply_gain(weight, weight_mask, out=weight, prod_map=products)
 
+        if "rebin_weight" in timestream.datasets:
+            timestream.rebin_weight[:] *= weight_mask
```
ssiegelx, line 2128: I think here `weight_mask` is `(nfreq, ninputs, ntime)` if dealing with un-stacked visibilities, or `(nfreq, 1, ntime)` if dealing with stacked visibilities, so that it can broadcast against the `weight` dataset.  So we need to remove the input/singleton axis when multiplying by `rebin_weight`.

PR #243, opened 2024-05-09
Add a task to establish a database connection

ljgray:
There have been a lot of issues with failed connections to chimedb from cedar recently. This PR adds a task to establish an initial connection with the database given some configurable timeout and number of attempts.

Depends on chime-experiment/chimedb#42 and chime-experiment/ch_util#69

### DIFF
--- a/ch_pipeline/core/dataquery.py
+++ b/ch_pipeline/core/dataquery.py
@@ -65,6 +65,35 @@ def _force_list(val) -> list:
         return [val]
 
 
+class ConnectDatabase(task.MPILoggedTask):
+    """Establish an initial connection to the chimedb database.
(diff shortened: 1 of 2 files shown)


ketiltrout (approved):
Really would be nicer if there was a pythonic way of specifying a timeout, but this is fine I suppose.

ljgray:
> Really would be nicer if there was a pythonic way of specifying a timeout, but this is fine I suppose.

I agree. I think that's what I had tried to do with the previous `ntries` argument everywhere, it just wound up being messy. I suppose I could pass it as an argument through to `.connect_database()`, but it will involve some modification of the chimedb code

PR #244, opened 2024-05-10
feat(flagging): add pipeline task for rainfall flagging

sjforeman:
New pipeline task to mask time samples after rainy periods, intended to replace the usage of `rain1mm` and `rain1mm_sd` flags from the database. The task computes the accumulated rainfall N hours prior to each time sample in a timestream or sidereal stream, and sets the weights to zero if the rainfall total exceeds some threshold (in mm). The user can specify N and the threshold.

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -9,7 +9,7 @@ from typing import Union
 import numpy as np
 
 from caput import mpiutil, mpiarray, memh5, config, pipeline, tod
-from ch_util import rfi, data_quality, tools, ephemeris, cal_utils, andata
+from ch_util import rfi, data_quality, tools, ephemeris, cal_utils, andata, finder
 from chimedb import dataflag as df
 from chimedb.core import connect as connect_database
 
@@ -18,6 +18,7 @@ from draco.core import task, io
 from draco.core import containers as dcontainers
 
 from ..core import containers
+from ..core.dataquery import _DEFAULT_NODE_SPOOF
 
 
 class RFIFilter(task.SingleTask):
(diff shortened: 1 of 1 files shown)


ch_pipeline/analysis/flagging.py
```diff
@@ -2260,3 +2261,195 @@ def process(self, data):
...
+    node_spoof : dictionary
+        Host and directory for finding weather data.
+        Default: {'cedar_online': '/project/rpp-krs/chime/chime_online/'}
+    """
+
+    accumulation_time = config.Property(proptype=float, default=30.0)
+    threshold = config.Property(proptype=float, default=1.0)
+    node_spoof = config.Property(proptype=dict, default=_DEFAULT_NODE_SPOOF)
+
+    def process(self, timestream):
```
ljgray, line 2401: Extremely nitpicky, but since this task can accept a sidereal stream as well as a time stream, using the variable name `timestream` could be confusing. Probably better to just use `datastream`, `data`, or something like that

ch_pipeline/analysis/flagging.py
```diff
@@ -2260,3 +2261,195 @@ def process(self, data):
...
+        # Compute mask corresponding to times when rainfall is below threshold
+        rainfall_mask = rainfall < self.threshold
+
+        # Multiply weights by mask
+        weight = timestream.weight[:]
+        weight.local_array[:] *= rainfall_mask[np.newaxis, np.newaxis, :]
+
+        # Report how much data has been flagged due to rainfall
+        self.log.info(
+            "%0.2f percent of data was flagged due to rainfall."
```
ljgray, line 2451:
```suggestion
            f"{100.0 * (1.0 - np.sum(rainfall_mask) / Len(rainfall_mask):.2f} percent of data was flagged due to rainfall."
```
Linter isn't enforced in `ch_pipeline` GitHub CI yet, but it will end up complaining about older-style strings eventually

ljgray (dismissed):
Looks good overall. A couple of minor comments

ljgray (approved):

PR #246, opened 2024-05-14
feat(flagging): add chime version of RFIMaskChisqHighDelay

ssiegelx:
Adds the appropriate hooks for a time mask based on daytime and source transits.  Also adds B0329+54 to the list of sources where we should use the MAD algorithm for RFISensitivityMask.

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -181,7 +181,7 @@ class RFIMaskChisqHighDelay(dflagging.RFIMaskChisqHighDelay):
         mask = np.zeros((freq.size, times.size), dtype=bool)
 
         for b_ in body:
-            mask |= transit_flag(b_, times, freq=freq, nsigma=self.transit_width)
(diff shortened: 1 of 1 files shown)


ssiegelx:
Depends on [draco #268](https://github.com/radiocosmology/draco/pull/268) for base class.

ch_pipeline/analysis/flagging.py
```diff
@@ -1246,29 +1309,35 @@ def daytime_flag(time):
     return flag
 
 
-def transit_flag(body, time, nsigma=2.0):
+def transit_flag(body, time, freq=400.0, nsigma=2.0):
```
ljgray, line 1312: `freq` argument should probably be after `nsigma` so as not to break any existing calls to this function which provide `nsigma` as a positional argument rather than a keyword

ljgray (dismissed):
This one looks good, other than one minor comment

ch_pipeline/analysis/flagging.py
ssiegelx: Good catch, thanks!

PR #247, opened 2024-05-16
feat(hfb): Task for on-off differencing

arashcosmology:


ch_pipeline/hfb/analysis.py
```diff
@@ -351,6 +351,104 @@ def process(self, minuend, subtrahend):
...
+
+    Used for flattening sub-frequency band shape by differencing on-source and
+    off-source data.
+    """
+
+    offset = config.Property(proptype=int, default=5)
+    nsamples = config.Property(proptype=int, default=5)
+
+    def process(self, stream):
+        """Take difference and place in container.
```
rikvl, line 365: It seems some bits of this docstring still need to be updated

ch_pipeline/hfb/analysis.py
```diff
@@ -351,6 +351,104 @@ def process(self, minuend, subtrahend):
...
+
+        Parameters
+        ----------
+        stream : HFBHighResRingMap
+            Container with stacked HFB data and weights.
```
rikvl, line 370: I think this line is maybe also from a previous task
```
            Container with stacked HFB data and weights.
```

ch_pipeline/hfb/analysis.py
```diff
@@ -351,6 +351,104 @@ def process(self, minuend, subtrahend):
...
+        # Take the weighted sum of off-source data (numerator of the
+        # weighted average). To do that, convolve the kernel above
+        # with the data
+        ker_fft = np.fft.fft(kernel)
+        dat_fft = np.fft.fft(data * weight, axis=2)
+        sum_data = np.fft.ifft(
+            dat_fft * ker_fft[np.newaxis, np.newaxis, :, np.newaxis], axis=2
+        ).real
+
+        # Then, convolve the kernel above  with the weights to find the sum
```
rikvl, line 402:
```suggestion
        # Then, convolve the kernel above with the weights to find the sum
```

ch_pipeline/hfb/analysis.py
```diff
@@ -351,6 +351,104 @@ def process(self, minuend, subtrahend):
...
+        # of off-source weights (denominator of the weighted average)
+        weight_fft = np.fft.fft(weight, axis=2)
+
+        # n is the sum of off-source weights over ra axis
+        n = np.fft.ifft(
```
rikvl, line 407: Not sure, but maybe it would be nice to call this variable `sum_weight`, since it's quite similar to `sum_data` above? Maybe I don't understand it well enough though

ch_pipeline/hfb/analysis.py
arashcosmology: Yes, it doesn't necessarily need to be stacked! `HFBHighResRingMap` is a container for HFB data and weights with a single high resolution frequency axis.

ch_pipeline/hfb/analysis.py
```diff
@@ -351,6 +351,104 @@ def process(self, minuend, subtrahend):
...
+        # value for n. This number goes to the denominator of weighted mean
+        # and makes the weighted mean very large. To avoid such numerical
+        # error (10^-16), zero n whereever weight is zero.
+        n[weight == 0] = 0
+
+        # Weighted average of off-source data
+        off = sum_data * tools.invert_no_zero(n)
+
+        # And on-off difference is:
+        on_off = data * weight * tools.invert_no_zero(weight) - off
```
rikvl, line 421:
```suggestion

        # Weighted average of on-source data
        on = data * weight * tools.invert_no_zero(weight)

        # And on-off difference is:
        on_off_diff = on - off
```

rikvl (changes requested):
This looks great already Arash. I have a few small comments, mostly on the "presentation" of the code

rikvl (approved):
Alrighty. If you're happy with it, I think you can do a "squash merge" (it should be one of the options in the drop-down menu at the bottom of your PR

PR #249, opened 2024-06-11
fix(hfb.containers): Add width axis and amplitude dataset to `HFBSearchResult`

rikvl:
Added an axis for the trial widths of the absorption features that we search for.
- The axis is called `width`, hopefully that understandable enough.
- I put this at the front of the axis order.
- I left the `el` axis to be the distributed axis, since we may not search over more widths than we use nodes in a given run.


rikvl:
Also added the `amplitude` dataset

arashcosmology (approved):

PR #254, opened 2024-06-25
fix(hfb.analysis): change distributed axis to el

arashcosmology:
Stacked data arriving at the on-off differencing task have `freq` axis as the distributed axis, while the container for on-off differencing task is distributed over `el` axis. I changed the distributed axis for on-off differencing task to `el` to be consistent with its container.


rikvl:
This looks good Arash, but I think you accidentally included the work-in-progress `HFBSearch` task in your final commit.

PR #255, opened 2024-06-26
Include slurm files in failed job check

ljgray:
Normally, the full log output is written to the `jobout.log` file. Occasionally, when a job fails because it reaches its time limit, the slurm message ("Job cancelled due to timeout") gets written to the slurm output file instead. Annoyingly, this doesn't seem to happen consistently. 

This change will check both the standard jobout file *and* the slurm output file.

Also, ensure that recent days are checked first in the `update_files` method of the daily processing. *NOTE*: this will have to be deployed if `auto-copy-to-online` script is removed @ketiltrout

### DIFF
--- a/ch_pipeline/processing/base.py
+++ b/ch_pipeline/processing/base.py
@@ -625,25 +625,32 @@ def classify_failed(
     failed = {k: [] for k in list(patterns.keys())}
 
     for tag in tags:
-        file = Path(dir) / tag / "job" / "jobout.log"
-        if not file.is_file():
-            continue
-        # Get the end of the file. Assume an average of 100 characters
-        # per line.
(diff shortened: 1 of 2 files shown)


ketiltrout (approved):
I _would_ like to move to a model where the daily pipeline is responsible for pulling new data online, but it doesn't have to be done immediately.  Would be necessary on the time scales of getting `cfm` working.

ljgray:
> I _would_ like to move to a model where the daily pipeline is responsible for pulling new data online, but it doesn't have to be done immediately. Would be necessary on the time scales of getting `cfm` working.

Ok. Whenever we want to deploy it I can just update the pipeline module on Cedar, which won't actually affect the rev_08 venv anyway.

PR #264, opened 2024-07-31
build: Change build to use pyproject

ketiltrout:
There was a `cython` requirement here which I've removed: there's no cython code in `ch_pipeline` and it appears to have been added here to help `cora` run its build system, which isn't the right way to do it.


.github/workflows/main.yml
```diff
@@ -30,28 +30,25 @@ jobs:
...
-        pip install cython
-        pip install --use-deprecated=legacy-resolver -r requirements.txt --no-binary bitshuffle
-        pip install --use-deprecated=legacy-resolver -r doc/requirements.txt --no-binary bitshuffle
-
     - name: Install ch_pipeline
-      run: pip install -e .
+      run: pip install -e . --no-binary bitshuffle
+
+    - name: Install sphinx
+      run: pip install -r doc/requirements.txt --no-binary bitshuffle
```
ljgray, line 51: Could be moved to docs section in optional-dependencies

ljgray (approved):

PR #231, opened 2024-01-17
Ruff linting

ljgray:
Adding ruff linting here. See https://github.com/radiocosmology/draco/pull/243

PR #266, opened 2024-08-02
fix: Use ch_ephem directly

ketiltrout:
This also resolves the use of a number of private symbols formerly imported from `ch_util.tools` which have been deleted from there.


ljgray:
@ketiltrout #231 is failing because of the missing symbols added here. I propose we merge this PR, rebase #231 with these fixes, then fix up any lingering ruff issues that show up.

ketiltrout:
That makes sense

ljgray (approved):

ljgray (approved):

PR #271, opened 2024-08-16
Check for tags which do not have data available anywhere

ljgray:
The part of the pipeline that checks for files to retrieve just looked at all the tags in the config and tried to fetch the upcoming ones that aren't online. The problem is that it didn't check to see if the tags it was trying to retrieve actually had _any_ data available anywhere, so it eventually gets stuck trying to request the same days over and over again, which will never end up online because there aren't any files.

This fix lets the pipeline first check whether a CSD has the minimum amount of data needed to run _anywhere_, not just on cedar_online.

### DIFF
--- a/ch_pipeline/processing/daily.py
+++ b/ch_pipeline/processing/daily.py
@@ -968,27 +968,30 @@ class DailyProcessing(base.ProcessingType):
 
         This includes any that currently exist or are in the job queue.
         """
-        # Get all desired csds from the config file
-        csds = self._all_tags
-        # Automatically exclude bad days
-        csds = [csd for csd in csds if csd not in self._exclude]
-        csds_sorted = sorted(csds)
-        # Get a list of CSDS whose files are entirely available online
-        csds_to_run = available_csds(
-            csds_sorted,
-            pad=self._padding,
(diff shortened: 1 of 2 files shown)


ch_pipeline/processing/daily.py
```diff
@@ -968,28 +968,32 @@ def _available_tags(self) -> list:
...
 
     @property
-    def _all_tags(self) -> list:
+    def _config_tags(self) -> list:
         """Return all tags desired from the config."""
         return unique_ordered(
             csd for i in self._intervals for csd in expand_csd_range(*i)
         )
 
+    def runnable_tags(self, require_online=False) -> list:
```
ketiltrout, line 982: It's odd that `require_online` defaults to False here, but in `available_csds` below, it defaults to True.  I'd make the defaults the same (one way or the other).

ketiltrout (approved):
This is great.  I see now why this was so difficult to sort out.  It's pretty subtle.

ch_pipeline/processing/daily.py
ljgray: Ah good call

PR #272, opened 2024-08-19
fix(beam): properly re-center HA grid after unwrapping long transits

sareda26:
Fixes an issue where the HA coordinate of very long transits (~180 degrees) would not be properly re-centered about HA=0, causing the holography regridder (`ch_pipeline.analysis.beam.TransitRegridder`) to silently fail, returning a stream of 0s in the output. See the attached plot for an example of the old and new behavior for a long transit of Tau A (RA ~ 83) which wraps at 360 degrees. 

image

### DIFF
--- a/ch_pipeline/analysis/beam.py
+++ b/ch_pipeline/analysis/beam.py
@@ -1278,9 +1278,9 @@ def unwrap_lha(lsa, src_ra):
     lsa += start_lsa
     # subtract source RA
     return np.where(
-        np.abs(lsa - src_ra) < np.abs(lsa - src_ra + 360.0),
+        np.abs(lsa - src_ra) < np.abs(lsa - (src_ra + 360.0)),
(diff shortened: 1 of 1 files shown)


ch_pipeline/analysis/beam.py
```diff
@@ -1278,9 +1278,9 @@ def unwrap_lha(lsa, src_ra):
     lsa += start_lsa
     # subtract source RA
     return np.where(
-        np.abs(lsa - src_ra) < np.abs(lsa - src_ra + 360.0),
+        np.abs(lsa - src_ra) < np.abs(lsa - (src_ra + 360.0)),
```
ljgray, line 1281: I don't think that the brackets here are actually necessary

ljgray (approved):

ssiegelx (approved):

PR #276, opened 2024-09-13
ss/sidereal mean

ssiegelx:
This contains all of the modifications to ch_pipeline required to run the daily foreground filtering pipeline.  It introduces a new set of tasks to correct for gain errors, mask periods of poor data quality, and mask regions near bright sources in ringmaps.  It updates several tasks to work with HybridVisStreams.

PR #278, opened 2024-11-21
fix(analysis.flagging.MaskBrightSourcePixels): Handle HA across edge.

tristpinsm:
I noticed that HA that spanned over an edge of the RA range were not being flagged.

After this fix I'm still finding some odd behaviour, but I haven't figured out the mechanism: One source has it's track mask cut off just below (but not at) the NCP. You can see in this figure:
image


ssiegelx:
Thanks Tristan.  Regarding the source track that ends, is it just that the source is setting below the horizon?  Currently it will not mask pixels where telescope-z < 0.

ssiegelx:
The phase wrapping in hour angle has been corrected in PR #276 prior to merging into master.  I will close this PR.

PR #279, opened 2024-12-03
Check for duplicate jobs and cancel all but one of them

ljgray:
There have been a number of times where multiple jobs for the same CSD end up queued and running, which burns an excessive amount of compute resources. It seems to happen when Cedar starts to respond after being unresponsive (but not technically inaccessible) for a while.

### DIFF
--- a/ch_pipeline/processing/base.py
+++ b/ch_pipeline/processing/base.py
@@ -13,6 +13,7 @@ import re
 import subprocess as sp
 import tempfile
 import warnings
+from datetime import datetime
 from pathlib import Path
 from typing import ClassVar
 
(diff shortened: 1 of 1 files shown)


ch_pipeline/processing/base.py
```diff
@@ -735,6 +739,106 @@ def slurm_split(line):
...
+        existing = names.get(name)
+
+        if existing is None:
+            # First time we've seen this job name
+            names[name] = (id, status, runtime, subtime)
+            continue
+
+        if existing[1] == "R":
+            # Both jobs are running. Keep the one that
+            # started most recently
```
ketiltrout, line 782: Is the idea here the most recent copy of a job is overwriting the output of the older copy, so it's the only version that can now produce a completed output?

ketiltrout (approved):
This looks good.

I'm sure there's bound to be some race condition-y thing going on with the state of jobs in the slurm queue changing between query and cancel, but it's likely going to be rare and I don't see a reasonable way of getting around that.

I think you'll need to rebase and update some of this for 3.10 (you're introducing an `Optional`).

ch_pipeline/processing/base.py
ljgray: Yeah, that's the idea

ketiltrout: Makes sense.

ketiltrout (approved):

PR #283, opened 2025-02-07
Ar/offline telescope

sareda26:
This PR adds functionality to `ch_pipeline.core.telescope.CHIME` which allows the feed layout of a `CHIME` instance to be obtained from a file saved to this repository, circumventing the usual database query. 

Reading from the file is now the default behavior; this can be overridden and a database query used by setting `read_local_layout: False` when configuring from YAML in a pipeline job, or passing `read_local_layout=False` to `CHIME.from_layout` in an interactive setting.

### DIFF
--- a/ch_pipeline/core/telescope.py
+++ b/ch_pipeline/core/telescope.py
@@ -209,7 +209,10 @@ class CHIME(telescope.PolarisedTelescope):
             # If we're reading locally, use the pickled layouts file
             feeds = self._read_local_layout()
         else:
-            # If not, fetch feed layout from database
+            feeds = None
+
+        # If not, or if the file I/O failed, fetch feed layout from database
+        if feeds is None:
             feeds = tools.get_correlator_inputs(self.layout, self.correlator)
 
         if mpiutil.size > 1:
@@ -223,10 +226,10 @@ class CHIME(telescope.PolarisedTelescope):
     def _read_local_layout(self):
         """Load the telescope layout from a file saved to this repository.
 
-        Note that this function will fallback to a database query if the
-        requested layout is earlier than September 1, 2018. As with
-        :method:`CHIME._load_layout`, this routine should not be called
-        directly.
+        If the file I/O fails, or if a layout from before the earliest recorded
+        date is requested, this method returns `None`, triggering a fallback
+        database query. As with :method:`CHIME._load_layout`, this routine
+        should not be called directly.
         """
         # Do I/O, and resolve the layout, only on rank 0
         if mpiutil.rank == 0:
@@ -238,8 +241,16 @@ class CHIME(telescope.PolarisedTelescope):
 
             layout_path = files(telescope_files).joinpath("layouts.pkl")
 
-            with layout_path.open("rb") as layout_f:
-                layouts = pickle.load(layout_f)
+            try:
+                with layout_path.open("rb") as layout_f:
+                    layouts = pickle.load(layout_f)
+
+            except OSError as e:
+                logger.warning(
+                    f"Failed to load local layout: {e}. Will try to "
+                    "load from the database."
+                )
+                return None
 
             # Load layout start and end times in arrays for comparisons
             lay_start = np.array([lay["start"] for lay in layouts])
@@ -255,7 +266,7 @@ class CHIME(telescope.PolarisedTelescope):
                     f"after {lay_start[0].strftime('%B %d, %Y')}."
                 )
 
-                feeds = tools.get_correlator_inputs(self.layout, self.correlator)
+                feeds = None
 
             # ... or later than the latest
             elif self.layout > lay_start[-1]:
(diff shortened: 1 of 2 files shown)


ketiltrout:
I think you'll need to add something to `pyproject.toml` to ensure the data file gets installed with the package. Something like this:
```
[tool.setuptools.package-data]
"ch_pipeline.tools.telescope_files" = ["layouts.pkl"]
```

sareda26:
> I think you'll need to add something to `pyproject.toml` to ensure the data file gets installed with the package. Something like this:
> 
> ```
> [tool.setuptools.package-data]
> "ch_pipeline.tools.telescope_files" = ["layouts.pkl"]
> ```

Right, thanks! Although, should that second line read `"ch_pipeline.core.telescope_files" = ...`?

ketiltrout:
> Right, thanks! Although, should that second line read `"ch_pipeline.core.telescope_files" = ...`?

Yeah, that's right.  The left-hand side is whatever the module path is.

ch_pipeline/core/telescope.py
```diff
@@ -87,6 +93,7 @@ class CHIME(telescope.PolarisedTelescope):
     layout = config.Property(default=None)
```
ssiegelx, line 93: What would happen if layout is not provided?  Should we have a default value here?

ssiegelx (approved):
Great job, thanks @sareda26!

ch_pipeline/core/telescope.py
sareda26: Yeah, good point; this is something we may as well settle on now. 

`_load_layout` will raise an exception if `layout` is not already set, but in `_finalise_config`, the call to `_load_layout` is conditioned on whether `layout` has been set; so, it seems that you are just (silently) left with an instance with no feed information in this case, which I imagine will tend to result in a crash somewhere downstream anyway. So, we should probably either crash here to require the user to be explicit in setting the layout (which would also be consistent usage with requiring a `layout` argument when building an instance  with `CHIME.from_layout`), or for convenience, set a default. In the latter case, do we have a reason to not use the most recent layout as the default?

ssiegelx: Yeah I think I would probably issue a warning that the parameter has not been set and then default to the most recent layout available in the files.

ch_pipeline/core/telescope.py
```diff
@@ -203,13 +220,84 @@ def _load_layout(self):
...
+
+            # Get the path of the layout file
+            from importlib.resources import files
+
+            from . import telescope_files
+
+            layout_path = files(telescope_files).joinpath("layouts.pkl")
+
+            with layout_path.open("rb") as layout_f:
+                layouts = pickle.load(layout_f)
```
ketiltrout, line 242:
```suggestion
            try:
                with layout_path.open("rb") as layout_f:
                    layouts = pickle.load(layout_f)
            except OSError as e:
                logger.warning(f"Failed to load local layout: {e}.  Will try to load layout from the database")
                return None  # Or whatever
```

If you do implement the fallback, I also think it would be a good idea to wrap the actual I/O here in a `try ... except OSError` block to catch I/O errors, which you could then handle by returning the failure result, meaning the pipeline will seamlessly fall back on trying to load the layout from the database in this case.

ketiltrout (dismissed):
I think this is probably fine as-is.

However! if you want to make it more robust, it might be better to use a fallback onto database query if the local layout load fails.  This could, perhaps, mitigate I/O problems reading the layout from disk.

I think the easiest way to do this would be to do something like:
```
        if self.read_local_layout:
            # If we're reading locally, use the pickled layouts file
            feeds = self._read_local_layout()
        else:
            feeds = None  # Or some other value indicating load failure

        # If reading from the local file didn't return anything (perhaps because the
        # load was skipped), fetch feed layout from database
        if feeds is None:
            feeds = tools.get_correlator_inputs(self.layout, self.correlator)
```
where `self._read_local_layout` returns some sort of result indicating load failure which the calling function can then handle.  I've used `None` but I know you're already returning `None` from this function on higher MPI ranks.  You'd have to think about how to deal with that.  Maybe a different value than `None` could be used as the failure return value.

Then you could remove the `tools.get_correlator_inputs` call from `_read_local_layout` (on line 258) and just return this failure result (i.e. `None` or whatever it is) so that the calling function can handle the database load.

ch_pipeline/core/telescope.py
ketiltrout: This would also catch the case where the user has failed to properly install the `layouts.pkl` file.

sareda26:
Thanks @ketiltrout! Yeah I agree that this is better and it's nice to keep the database query in only one place. I have implemented basically what you have described, except that I kept `None` as the return value in case of failure and instead return an empty list on higher MPI ranks.

sjforeman (approved):
This works on my laptop, so it seems like it's good to go, unless there are any final modifications that have been recommended

ketiltrout (approved):
Very nice.

PR #284, opened 2025-02-24
Tasks to generate a CHIME rfi mask using HFB data

ljgray:
Developed by @yukari-u


ljgray:
Note for Yukari - I've just rebased your commits onto the `master` branch, and squashed them down into a smaller number of commits.

yukari-u (approved):

yukari-u:
To Rik and Arash, 
Since the HFBRadiometerRFIFlagging task contains duplicated code for calculating sensitivity values in HFB data, I plan to create an HFBSensitivity container and an HFBSensitivityCalculation class to eliminate code repetition in the future. Additionally, the current HFBRadiometerRFIFlagging task has a bug when computing sensitivity values, so the future update will be beneficial. For now, these commits focus on HFB-based RFI flagging.

PR #286, opened 2025-03-22
refactor(analysis): Use `object_coords` Observer method

ketiltrout:
This all still works as-is, but since Seth ported this function to `caput` as an `Observer` method (https://github.com/radiocosmology/caput/pull/290), let's use that directly.


ljgray (approved):

ssiegelx (approved):
Thanks @ketiltrout!

PR #288, opened 2025-03-25
Defining a function to compute sensitivity metric

yukari-u:
1) I newly defined a new function 'sensitivity_metric' to compute sensitivity metric.
2) I fixed the bug in 'HFBRadiometerRFIFlagging' and updated it to use the function above.
3) I renamed 'HFBSensitivityFlagging' to 'HFBDirectionalRFIFlagging' and updated it to use the function above.

### DIFF
--- a/ch_pipeline/hfb/containers.py
+++ b/ch_pipeline/hfb/containers.py
@@ -569,7 +569,7 @@ class HFBSearchResult(HFBRingMapBase, HFBHighResContainer):
         return self.datasets["amplitude"]
 
 
-class HFBSensitivityMask(FreqContainer, TODContainer):
+class HFBDirectionalRFIMask(FreqContainer, TODContainer):
     """Container for an RFI mask based on HFB data.
 
     This RFI mask takes advantage of the high-frequency resolution of HFB data.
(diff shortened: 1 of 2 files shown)


rikvl (dismissed):
This looks very good.

I have one comment, but it doesn't really relate to this PR. You can decide if you want to make that change or not.

For the commit message, have a look at the [Commit Message Guidelines](https://github.com/chime-experiment/Pipeline/blob/master/CONTRIBUTING.md#commit). This included both a refactor and a bug fix. I think the commit message should be something like
```
fix(hfb.flagging): fix incorrect frac_lost
```

rikvl (dismissed):
This is fine if you remove the superfluous line.

Regarding the commit messages: I think this last commit would constitute a "fix" rather than a "refactor". When the PR is ready to merge into `master`, you should do a "Squash and merge". At that point you will be able to write a new commit message for the entire PR (the different commits will be "squashed" into a single commit)

rikvl (approved):

arashcosmology:
Looks good!

arashcosmology (approved):

PR #292, opened 2025-05-04
feat(flagging): support for tapering maps around the location of sources

ssiegelx:
* Refactor catalog-based masking code.
* Create TaperSourcePixelsFromCatalog and TaperSourceTracksFromCatalog,
which generate a smooth transition from 1 to 0 around the location
of bright soruces.
* Add ability to mask a subset of frequencies based on source redshift.
* Add support for a user-specified horizon limit.

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -18,6 +18,7 @@ from draco.analysis import flagging as dflagging
 from draco.analysis.ringmapmaker import find_grid_indices
 from draco.core import containers as dcontainers
 from draco.core import io, task
+from scipy.spatial import KDTree
 
 from ..core import containers
 from ..core.dataquery import _DEFAULT_NODE_SPOOF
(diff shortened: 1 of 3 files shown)


ssiegelx:
I think ideally the external horizon limits should be incorporated into the telescope class. However, this PR and the corresponding PR in draco implement a solution tailored specifically to the catalog-based mask generation, so that we can produce appropriate masks for the power spectrum analysis.

ljgray (approved):

ch_pipeline/analysis/flagging.py
```diff
@@ -2841,7 +2842,206 @@ def search_grid(xeval, window, x, wrap=False):
...
+            the rest-frame frequency of the 21 cm line.
+        src_freq_err : np.ndarray[nsource,]
+            Uncertainty in source frequency in MHz, propagated from redshift errors.
+        """
+        from cora.util import units
+
+        z = self.catalog["redshift"]["z"][:]
+        zerr = self.catalog["redshift"]["z_error"][:]
+
+        src_freq = units.nu21 / (1.0 + z)
```
Arnab-half-blood-prince, line 2894: this is fine, but if one wants to mask the OH-line, then this freq, which is set to 21cm line, needs to be changed. Maybe later on, we include a parameter, like whether OH or HI-21cm line freq to use.

ch_pipeline/analysis/flagging.py
```diff
@@ -3071,37 +3241,142 @@ def get_source_coordinates(self, data):
...
+            b = scipy.constants.c / (freq[:, np.newaxis] * 1e6 * self.min_ysep)
+            src_yalias = src_y + (1.0 - 2.0 * (src_y > 0.0)) * b  # nfreq, nsource
+
+        # Get the size of the window
+        nu = np.full((nfreq, 1), min_freq) if self.common_freq else freq[:, np.newaxis]
+        wavelength = scipy.constants.c / (nu * 1e6)
+
+        ## In the sin(za) direction, we use the synthesized beam.
+        ## 0.85 * wavelength / max_ysep gives the FWHM for a natural weighting scheme.
+        sigma_y = 0.85 * wavelength / (self.max_ysep * 2.35482)
```
Arnab-half-blood-prince, line 3291: maybe also mention the description of the 2.35482 factor ...

Arnab-half-blood-prince (dismissed):

ch_pipeline/analysis/flagging.py
ssiegelx: I've started saving the frequency in addition to the redshift in the SpectralLineCatalog and using that directly here.

I agree that in the future we could have a config property specifying what line to use for external catalogs.

Arnab-half-blood-prince (approved):

ljgray (approved):

ljgray:
> I think ideally the external horizon limits should be incorporated into the telescope class. However, this PR and the corresponding PR in draco implement a solution tailored specifically to the catalog-based mask generation, so that we can produce appropriate masks for the power spectrum analysis.

I had started doing that ages ago in this branch https://github.com/chime-experiment/ch_pipeline/tree/ljg/horizon-beam, so we could revive it at some point,

PR #293, opened 2025-05-04
feat(spectral_lines): task to identify candidate spectral lines

ssiegelx:
Takes all outlier excursions in a RingMap, clusters them based on their proximity using some normalized 3D distance metric, characterizes each cluster, and populates a new SpectralLineCatalog container with the results.

### DIFF
--- a/ch_pipeline/analysis/spectral_lines.py
+++ b/ch_pipeline/analysis/spectral_lines.py
@@ -225,6 +225,8 @@ class FindSpectralLines(task.SingleTask):
             # Save all results for this source to the output container
             out["position"]["ra"][cc] = src_body.ra._degrees
             out["position"]["dec"][cc] = src_body.dec._degrees
+            out["frequency"]["freq"][cc] = src_freq
+            out["frequency"]["freq_error"][cc] = dfreq
             out["redshift"]["z"][cc] = units.nu21 / src_freq - 1.0
             out["redshift"]["z_error"][cc] = units.nu21 * dfreq / src_freq**2
             out["flux"]["continuum"][cc] = 0.0
--- a/ch_pipeline/analysis/spectral_lines.py
+++ b/ch_pipeline/analysis/spectral_lines.py
@@ -1,0 +1,253 @@
+"""Find and characterize spectral lines."""
+
+import numpy as np
+import scipy.constants
+from caput import config, time
+from ch_util import cal_utils
+from cora.util import units
+from draco.analysis.ringmapmaker import find_grid_indices
+from draco.core import io, task
+from draco.util import tools
+from scipy.cluster.hierarchy import fcluster, linkage
+from scipy.spatial.distance import pdist
+from skyfield.units import Angle
+
+from ..core import containers
+
+
+class FindSpectralLines(task.SingleTask):
+    """Identify candidate spectral line features in continuum-filtered RingMap data.
+
+    This task detects statistically significant spectral excursions, either positive
+    (emission) or negative (absorption), that are consistent across polarizations.
+    Outliers are clustered in (frequency, ra, el) space using a distance metric
+    scaled to the instrument's spectral and angular resolution. Each cluster is
+    interpreted as a potential spectral line and recorded in a `SpectralLineCatalog`.
+
+    Attributes
+    ----------
+    delay_cutoff : float
+        Delay cutoff in microseconds used during the continuum filtering step.
+        Defines the resolution in frequency and sets the scale for associating
+        outliers into candidate features.
+    nsigma : float
+        Maximum allowed clustering distance (in units of normalized coordinate space)
+        between outliers to be grouped as a single spectral line.
+    """
+
+    delay_cutoff = config.Property(proptype=float, default=0.20)
+    nsigma = config.Property(proptype=float, default=2.0)
+
+    def setup(self, manager):
+        """Set up the telescope instance.
+
+        Parameters
+        ----------
+        manager : ProductManager
+            Product manager from which a telescope instance is extracted.
+        """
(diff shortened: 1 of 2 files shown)


ch_pipeline/analysis/spectral_lines.py
```diff
@@ -0,0 +1,253 @@
...
+                local_s2n = 0.0
+                local_imax = -1
+
+            # Take maximum over all ranks
+            local_info = [local_max, local_s2n, local_imax, local_npix]
+            all_info = self.comm.allgather(local_info)
+
+            global_max, global_s2n, global_imax, _ = max(
+                all_info, key=lambda x: np.abs(x[0])
+            )
```
rikvl, line 211: See previous comment

ch_pipeline/analysis/spectral_lines.py
```diff
@@ -0,0 +1,253 @@
...
+                # Need to convert back to local indices to extract map values
+                local_ind = tuple(
+                    [ind[this_rank] - off for ind, off in zip(index, offset_local)]
+                )
+
+                map_values = rm[local_ind]
+                s2n_values = np.abs(map_values) * np.sqrt(wm[local_ind])
+
+                # Identify the maximum pixel within the local part of this cluster
+                imax = np.argmax(np.abs(map_values))
```
rikvl, line 195: I'm wondering if you'd want to instead select the pixel with the maximum SNR, since the pixel with the maximum absolute value could also have that value due to being very noisy. Not sure if I understood correctly. If what I'm thinking doesn't make sense, could you explain?

rikvl (approved):
This looks good. I have a few minor comments and a question.

Something I'm wondering about: if a spectral feature is located close to the edge of a frequency bin, it can affect two adjacent frequency channels. If the `delay_cutoff` and `nsigma` attributes are such that the threshold distance in the frequency dimension is smaller than the frequency resolution (390 kHz), would the clustering algorithm still group these two adjacent frequency bins together? Or is that just something the user needs to pay attention to when choosing values for those attributes?

Somewhat related (maybe not for this PR): there are mechanisms that can broaden or spread spectral features in the frequency dimension, such as the Doppler effect due to Earth's motion (max amplitude of ~80 kHz). Should there be a way to consider such mechanisms for clustering purposes, in addition to the spectral resolution set by `delay_cutoff`?

ch_pipeline/analysis/spectral_lines.py
```diff
@@ -0,0 +1,253 @@
...
+            max_ysep = ringmap.attrs["beamform_ns_nsmax"]
+            if ringmap.attrs["beamform_ns_scaled"]:
+                freq_eval = ringmap.attrs["beamform_ns_freqmin"]
+
+        wavelength = scipy.constants.c / (freq_eval * 1e6)
+
+        delta = np.zeros((noutlier, 3), dtype=float)
+        delta[:, 0] = 1.0 / (2.35482 * self.delay_cutoff)
+        delta[:, 1] = cal_utils.guess_fwhm(freq_eval, pol="X", dec=src_dec, sigma=True)
+        delta[:, 2] = 0.85 * wavelength / (2.35482 * max_ysep)
```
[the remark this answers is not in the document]
ssiegelx: I now give reasons for each choice of normalization.

ch_pipeline/analysis/spectral_lines.py
```diff
@@ -0,0 +1,253 @@
...
+
+        coord = raw_coord / delta
+
+        # Compute pairwise distances and cluster
+        dists = pdist(coord)  # pairwise distances in rescaled space
+        linkage_matrix = linkage(dists, method="single")  # or 'complete', 'average'
```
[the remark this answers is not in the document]
ssiegelx: I've actually not tried the other options, but it would be worth exploring at some point.  For now, I've made this a config property that the user can change.

ch_pipeline/analysis/spectral_lines.py
ssiegelx: Yeah that make sense.  I've updated the code to use the pixel with the maximum signal to noise.

ssiegelx:
> Something I'm wondering about: if a spectral feature is located close to the edge of a frequency bin, it can affect two adjacent frequency channels. If the delay_cutoff and nsigma attributes are such that the threshold distance in the frequency dimension is smaller than the frequency resolution (390 kHz), would the clustering algorithm still group these two adjacent frequency bins together? Or is that just something the user needs to pay attention to when choosing values for those attributes?

I think this is something the user needs to pay attention to by making an informed choice of the `delay_cutoff` parameter. In practice, the current value corresponds to a maximum clustering distance of about 4 MHz, which is much larger than the 390 kHz frequency resolution. This value is set based on the delay cutoff we use for continuum filtering. It’s true that if you have a narrow spectral line, it will naturally leak into neighboring bins up to ~4 MHz away because of the filter, so clustering over that scale is expected.

> Somewhat related (maybe not for this PR): there are mechanisms that can broaden or spread spectral features in the frequency dimension, such as the Doppler effect due to Earth's motion (max amplitude of ~80 kHz). Should there be a way to consider such mechanisms for clustering purposes, in addition to the spectral resolution set by delay_cutoff?

I think these effects are small compared to the broadening introduced by the continuum filtering.

There are definitely some improvements that could be made to the clustering algorithm, but for now the source density is low enough that the current approach seems to work reasonably well.

rikvl (approved):
This looks great, thanks for addressing all the comments!

One thing you might still consider is to list the default values for the attributes in the docstring (`delay_cutoff = 0.2`, `linkage_method = "single"`, `nsigma=2.0`)

And my apologies for taking so long to get back to this. It fell off my radar. Don't hesitate to send me a reminder if this happens

PR #294, opened 2025-05-06
feat(hfb/container): add compression option to HFBDirectionalRFIMask

yukari-u:
Add compression option to the frac_rfi dataset of HFBDirectionalRFIMask

### DIFF
--- a/ch_pipeline/hfb/containers.py
+++ b/ch_pipeline/hfb/containers.py
@@ -6,19 +6,15 @@ from typing import ClassVar
 import numpy as np
 from caput import memh5, tod
 from ch_util import andata
-from draco.core.containers import DataWeightContainer, SiderealContainer, TODContainer
+from draco.core.containers import (
+    COMPRESSION,
+    COMPRESSION_OPTS,
+    DataWeightContainer,
+    SiderealContainer,
+    TODContainer,
+)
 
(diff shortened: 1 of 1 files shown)


ch_pipeline/hfb/containers.py
```diff
@@ -10,6 +10,16 @@
...
+# Try to import bitshuffle to set the default compression options
+try:
+    import bitshuffle.h5
+
+    COMPRESSION = bitshuffle.h5.H5FILTER
+    COMPRESSION_OPTS = (0, bitshuffle.h5.H5_COMPRESS_LZ4)
+except ImportError:
+    COMPRESSION = None
+    COMPRESSION_OPTS = None
+
```
ljgray, line 22:
```suggestion
from draco.core.containers import COMPRESSION, COMPRESSION_OPTS
```

Since we don't actually include `bitshuffle` as a dependency in `ch_pipeline`, and we already import containers from `draco`, we can just import these objects directly as well. It also means that we only need to maintain this import logic in one place.

ljgray (approved):
This looks great. I just have one comment about the imports.

ljgray (approved):

rikvl (approved):

arashcosmology (approved):

PR #302, opened 2025-08-07
fix(hfb): Fixing a bug and updating default values

yukari-u:
- Fixed `self.bitmap.get(std_key)` to `self.bitmap.get(str(std_key))`
- Updated default `std` values in `HFBDirectionalRFIFlagging` to [0.25, 0.30, 0.35, 0.50]

### DIFF
--- a/ch_pipeline/hfb/containers.py
+++ b/ch_pipeline/hfb/containers.py
@@ -4,8 +4,8 @@ from functools import cached_property
 from typing import ClassVar
 
 import numpy as np
+from caput import memdata
 from caput.containers import tod
-from caput.memdata import _memh5
 from ch_util import andata
 from draco.core.containers import (
     COMPRESSION,
@@ -28,7 +28,7 @@ class HFBContainer(DataWeightContainer):
     _weight_dset_name = None  # Leave as None as this could potentially change location
 
     @property
-    def hfb(self) -> _memh5.MemDataset:
+    def hfb(self) -> memdata.MemDataset:
         """Convenience access to the main hfb dataset."""
         if "hfb" in self.datasets:
             return self.datasets["hfb"]
@@ -36,7 +36,7 @@ class HFBContainer(DataWeightContainer):
         raise KeyError("Dataset 'hfb' not initialised.")
 
     @property
-    def weight(self) -> _memh5.MemDataset:
+    def weight(self) -> memdata.MemDataset:
         """The inverse variance weight dataset."""
         if "weight" in self:
             weight = self["weight"]
@@ -49,7 +49,7 @@ class HFBContainer(DataWeightContainer):
         return weight
 
     @property
-    def nsample(self) -> _memh5.MemDataset:
+    def nsample(self) -> memdata.MemDataset:
         """Get the nsample dataset (number of non-zero samples) if it exists."""
         if "nsample" in self.datasets:
             return self.datasets["nsample"]
@@ -220,7 +220,7 @@ class HFBReader(tod.TODReader):
 
         Parameters
         ----------
-        out_group : `h5py.Group`, hdf5 filename or `memh5.Group`
+        out_group : `h5py.Group`, hdf5 filename or `memdata.Group`
             Underlying hdf5 like container that will store the data for the
             BaseData instance.
 
@@ -621,7 +621,7 @@ class HFBDirectionalRFIMaskBitmap(FreqContainer, TODContainer):
 
             # If duplicates exist, print them and proceed with unique values only
             if duplicates:
-                print(
+                self.log.warning(
                     f"Duplicate sigma_key values provided (within atol=0.001): {duplicates}. "
                     f"Using unique values only: {sigma_unique}."
                 )
@@ -698,9 +698,9 @@ class HFBDirectionalRFIMaskBitmap(FreqContainer, TODContainer):
             if np.isclose(float(k), float(sigma_key), rtol=0, atol=0.001):
                 offset = v
                 break
-        if offset is None:
+        else:
             raise KeyError(
-                f"Invalid sigma_key '{sigma_key}'. Must be one of {list(self.bitmap.keys())}."
+                f"Invalid sigma_key {sigma_key}. Must be one of {list(self.bitmap.keys())}."
             )
 
         return ((self.subfreq_rfi[:] >> (8 * offset)) & 0xFF).astype(np.uint8)
@@ -717,10 +717,11 @@ class HFBDirectionalRFIMaskBitmap(FreqContainer, TODContainer):
             if np.isclose(float(k), float(sigma_key), rtol=0, atol=0.001):
                 offset = v
                 break
-        if offset is None:
+        else:
             raise KeyError(
-                f"Invalid sigma_key '{sigma_key}'. Must be one of {list(self.bitmap.keys())}."
+                f"Invalid sigma_key {sigma_key}. Must be one of {list(self.bitmap.keys())}."
             )
+
         if np.any((values < 0) | (values > 128)):
             raise ValueError("Values must be in range 0 to 128.")
 
@@ -775,7 +776,7 @@ class HFBDirectionalRFIMaskBitmap(FreqContainer, TODContainer):
                 mask.sum(axis=-1) > remove_persistent_beamns_frac * ntime
             )  # persistent shape: (freq, beam_ns)
 
-            mask &= ~persistent[..., None]
+            mask &= ~persistent[..., np.newaxis]
 
         return mask
 


ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,66 @@ def process(self, rfimaskbitmap):
         return out
 
 
+class RFIMaskReduceBeamNS(task.SingleTask):
+    """Create RFIMask from HFBDirectionalRFIMaskBitmap.
+
+    This task takes an HFBDirectionalRFIMaskBitmap)freq, beam_ns, time), selects the RFI mask corresponding to a specified significance value used in the detection and subfrequency threshold, then reduce the 'beam_ns' axis to create a RFIMask container (freq, time).
```
ljgray:
```suggestion
    This task takes an HFBDirectionalRFIMaskBitmap(freq, beam_ns, time), selects the RFI mask corresponding to a specified significance value used in the detection and subfrequency threshold, then reduce the 'beam_ns' axis to create a RFIMask container (freq, time).
```

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,66 @@ def process(self, rfimaskbitmap):
...
+        reduced_mask = np.sum(mask, axis=beam_ns_axis) >= self.beam_ns_threshold
+
+        # Create an output container
+        output = RFIMask(freq=freq, time=time)
+
+        # The output RFI mask is not frequency distributed
+        arr = reduced_mask
+        arrdist = mpiarray.MPIArray.wrap(arr, axis=0)
+        final_mask = arrdist.allgather()
+        output.mask[:] = final_mask
```
ljgray:
```suggestion
        output.mask[:] = mpiarray.MPIArray.wrap(reduced_mask, axis=0).allgather()
```
This can just be reduced to a single line, which in some way is more readable because there aren't intermediate variables.

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,66 @@ def process(self, rfimaskbitmap):
...
+        mask = rfimaskbitmap.get_mask(self.sigma, self.subfreq_threshold)[:]
+        beam_ns_axis = list(rfimaskbitmap.subfreq_rfi.attrs["axis"]).index("beam_ns")
+        freq = rfimaskbitmap.freq[:]
+        time = rfimaskbitmap.time[:]
+
+        # Apply reduction condition
+        reduced_mask = np.sum(mask, axis=beam_ns_axis) >= self.beam_ns_threshold
+
+        # Create an output container
+        output = RFIMask(freq=freq, time=time)
```
ljgray:
```suggestion
        output = RFIMask(axes_from=rfimaskbitmap, attrs_from=rfimaskbitmap)
```
Since the frequency and time axes are identical, there's no need to explicitly extract them from the input container - the `axes_from` argument will do that for you.

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,66 @@ def process(self, rfimaskbitmap):
...
+        -------
+        out : draco.containers.RFIMask(freq, time)
+            Non beam_ns-specific RFI mask with axes (freq, time).
+
+        """
+        # Extract mask/frac and axes data
+        mask = rfimaskbitmap.get_mask(self.sigma, self.subfreq_threshold)[:]
+        beam_ns_axis = list(rfimaskbitmap.subfreq_rfi.attrs["axis"]).index("beam_ns")
+        freq = rfimaskbitmap.freq[:]
+        time = rfimaskbitmap.time[:]
```
ljgray:
```suggestion
```

See below

ljgray (approved):
Looks good! I have a couple of comments, but none of them affect the functionality of the code.

ch_pipeline/hfb/flagging.py
```diff
@@ -130,29 +130,29 @@ class HFBDirectionalRFIFlagging(task.SingleTask):
...
         This is to detect RFI in HFB sensitivity values. If it exceeds
-        1 + treshold * std (or std_avg), then it indicates RFI events. Default is 5.
+        1 + sigma * std, then it indicates RFI events. Default is 0.25.
     """
 
     beam_ew_id = config.Property(proptype=int, default=1)
-    std = config.Property(proptype=list, default=[0.25, 0.275, 0.30, 0.325])
-    sigma_threshold = config.Property(proptype=float, default=5)
+    sigma = config.Property(proptype=list, default=[4, 5, 6, 10])
+    std = config.Property(proptype=float, default=0.25)
```
rhaas80: is `std` a very descriptive name? "standard" could be anything. If it's related to standard-deviation then "standard" to me is the descriptive attribute of the "deviation".

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,61 @@ def process(self, rfimaskbitmap):
...
+    sigma : int
+        Specify the value of significance used in RFI detection to create a RFI mask.
+        Must match one of the sigma keys used when the HFBDirectionalRFIMaskBitmap was
+        created. Default is 5.
+    subfreq_threshold : int
+        Subfrequency threshold for decoding the bitmap. A sample is flagged if
+        the number of HFB subfrequency channels detecting RFI >= this threshold.
+        Default is 2.
+    beam_ns_threshold : int
+        This number determines the minimum number of detected RFI events along the beam_ns axis required for a data point
```
rhaas80: line wrapping (number of characters per line) is inconsistent. Fixable by black or something like that?

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,61 @@ def process(self, rfimaskbitmap):
...
+            beam_ns-specific RFI mask with axes (freq, beam_ns, time).
+
+        Returns
+        -------
+        out : draco.containers.RFIMask(freq, time)
+            Non beam_ns-specific RFI mask with axes (freq, time).
+
+        """
+        # Extract mask/frac and axes data
+        mask = rfimaskbitmap.get_mask(self.sigma, self.subfreq_threshold)[:]
```
rhaas80: not being familar with the framework: "[:]" actually does something (forces things being converted to a numpy array or so?)?

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,61 @@ def process(self, rfimaskbitmap):
...
+
+    This task takes an HFBDirectionalRFIMaskBitmap(freq, beam_ns, time), selects the RFI mask corresponding to a specified significance value used in the detection and subfrequency threshold, then reduce the 'beam_ns' axis to create a RFIMask container (freq, time).
+
+    Attributes
+    ----------
+    sigma : int
+        Specify the value of significance used in RFI detection to create a RFI mask.
+        Must match one of the sigma keys used when the HFBDirectionalRFIMaskBitmap was
+        created. Default is 5.
+    subfreq_threshold : int
```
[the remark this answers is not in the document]
yukari-u: This should be an integer because subfreq_threshold represents the number of subfrequency channels out of 128. Discrete

ch_pipeline/hfb/flagging.py
```diff
@@ -348,6 +345,61 @@ def process(self, rfimaskbitmap):
...
+        Subfrequency threshold for decoding the bitmap. A sample is flagged if
+        the number of HFB subfrequency channels detecting RFI >= this threshold.
+        Default is 2.
+    beam_ns_threshold : int
```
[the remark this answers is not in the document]
yukari-u: This should be an integer because beam_ns_threshold represents the number of beam channels in the north–south direction out of 256. Discrete again.

ch_pipeline/hfb/flagging.py
yukari-u: This code passed the Black check, but I’ll fix it manually. Thank you for pointing it out.

ch_pipeline/hfb/flagging.py
```diff
@@ -130,29 +130,29 @@ class HFBDirectionalRFIFlagging(task.SingleTask):
...
         These thresholds are used to compare against the sensitivity metric, and each one
         produces a separate mask encoding the number of HFB subfrequency channels flagged as RFI.
         The order of values determines their byte position: the first maps to bits 0-7, the second
-        to 8-15, and so on. Default is [0.25, 0.275, 0.30, 0.325].
-    sigma_threshold : float
+        to 8-15, and so on. Default is [4, 5, 6, 10].
+    std : float
         This is to detect RFI in HFB sensitivity values. If it exceeds
-        1 + treshold * std (or std_avg), then it indicates RFI events. Default is 5.
+        1 + sigma * std, then it indicates RFI events. Default is 0.25.
```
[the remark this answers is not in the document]
yukari-u: The threshold is computed for each sigma value in the list, so the default is [0.25, 0.275, 0.30, 0.325]. I’ll write it here to remind the reader. Thanks for the point.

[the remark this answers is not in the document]
yukari-u: I add below
```
    Notes
    -----
    The RFI detection threshold is evaluated separately for each value in the sigma
    list using the expression 1 + sigma * std. The default values (std = 0.25 and
    sigma = [4, 5, 6, 10]) correspond to effective thresholds of [2.0, 2.25, 2.5, 3.5].
    These values were chosen to provide multiple levels of sensitivity within a single packed mask.

```

ch_pipeline/hfb/flagging.py
```diff
@@ -189,19 +197,19 @@ def process(self, stream):
             beam_ns=stream.beam_ns[:],
             time=stream.time[:],
         )
-        out.attrs["bitmap"] = {std: i for i, std in enumerate(self.std[:])}
+        out.attrs["bitmap"] = {std: i for i, std in enumerate(self.sigma[:])}
```
rhaas80: `std` in loop

ch_pipeline/hfb/flagging.py
```diff
@@ -349,6 +361,73 @@ def process(self, rfimaskbitmap):
...
+    subfreq_threshold = config.Property(proptype=int, default=2)
+    beam_ns_threshold = config.Property(proptype=int, default=1)
+    remove_persistent_beamns_frac = config.Property(proptype=float, default=0.0)
+
+    def process(self, rfimaskbitmap):
+        """Produce a RFI mask.
+
+        Parameters
+        ----------
+        rfimaskbitmap : containers.HFBDirectionalRFIMaskBitmap
```
rhaas80: no `draco.` here. Should there be (can this even cause confusion)?

ch_pipeline/hfb/containers.py
```diff
@@ -602,23 +602,25 @@ class HFBDirectionalRFIMaskBitmap(FreqContainer, TODContainer):
...
+    def __init__(self, *args, sigma_key: list[float] = [], **kwargs):
         """Sets up the bitmap attribute in the packed 32-bit representation."""
         super().__init__(*args, **kwargs)
 
-        # If std_key is provided, validat and store mapping for decoding individual 8-bit RFI segments
-        if std_key:
-            if len(std_key) != 4:
+        # If sigma_key is provided, validate and store mapping for decoding individual 8-bit RFI segments
+        if sigma_key:
+            if len(sigma_key) != 4:
```
ljgray: (This is just a suggestion) - technically you don't actually *need* 4 keys here, you can just have a maximum of 4. If we only wanted to save two, for example, we could just allow the other 16 bits to be zeros.

ch_pipeline/hfb/flagging.py
```diff
@@ -349,6 +361,73 @@ def process(self, rfimaskbitmap):
...
+    remove_persistent_beamns_frac: float
+        If non-zero, remove beam_ns rows that are persistently flagged over time
+        (i.e., flagged for more than the specified fraction of samples) because such persistent
+        flagging is interpreted as an instrumental offset rather than genuine RFI.
+    """
+
+    sigma = config.Property(proptype=float, default=5)
+    subfreq_threshold = config.Property(proptype=int, default=2)
+    beam_ns_threshold = config.Property(proptype=int, default=1)
+    remove_persistent_beamns_frac = config.Property(proptype=float, default=0.0)
```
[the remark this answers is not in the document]
ljgray: Defaulting to `None` is possible, but it would just require the code to do an additional check wherever this is used.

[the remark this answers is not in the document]
yukari-u: Ruff check didn't like having `None` there. I stick with the magic number.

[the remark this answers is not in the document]
ljgray: Oh, I mean I'm surprised that `ruff` wouldn't accept that - did you get it to work with `None` as a default?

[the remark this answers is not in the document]
ljgray: Great!

ch_pipeline/hfb/containers.py
```diff
@@ -30,15 +28,15 @@ class HFBContainer(DataWeightContainer):
     _weight_dset_name = None  # Leave as None as this could potentially change location
 
     @property
-    def hfb(self) -> memdata.MemDataset:
+    def hfb(self) -> _memh5.MemDataset:
```
ljgray: These types shouldn't be changed - `memdata` is the public interface

ch_pipeline/hfb/containers.py
```diff
@@ -632,60 +661,65 @@ def beam_ns(self):
...
-
+        offset = None
+        for k, v in self.bitmap.items():
+            if np.isclose(float(k), float(sigma_key), rtol=0, atol=0.001):
+                offset = v
+                break
         if offset is None:
             raise KeyError(
-                f"Invalid std_key '{std_key}'. Must be one of {list(self.bitmap.keys())}."
+                f"Invalid sigma_key '{sigma_key}'. Must be one of {list(self.bitmap.keys())}."
```
ljgray:
```suggestion
        for k, v in self.bitmap.items():
            if np.isclose(float(k), float(sigma_key), rtol=0, atol=0.001):
                offset = v
                break
        else:
            raise KeyError(f"Invalid sigma_key '{sigma_key}'. Must be one of {list(self.bitmap.keys())}."
```
This is functionally the same thing, but takes advantage of python's `for; else` syntax, where the `else` branch runs if the `for` loop doesn't break.

PR #303, opened 2025-08-12
Update default cluster params for Fir

ljgray:
Updates basic config stuff for `fir`. I've been using this already. Also fixes the new ruff rule.

### DIFF
--- a/ch_pipeline/analysis/flagging.py
+++ b/ch_pipeline/analysis/flagging.py
@@ -145,7 +145,7 @@ class RFIFilter(task.SingleTask):
         return out
 
 
-class RFIStokesIMask(dflagging.RFIStokesIMask):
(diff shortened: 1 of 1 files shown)


ketiltrout:
Can you also update the name in the default node spoof https://github.com/chime-experiment/ch_pipeline/blob/master/ch_pipeline/core/dataquery.py#L57

ljgray:
> Can you also update the name in the default node spoof https://github.com/chime-experiment/ch_pipeline/blob/master/ch_pipeline/core/dataquery.py#L57
> 
> (Also in the docs on line 108)

Done. Also updated a couple of instances of paths using `rpp-krs`

ketiltrout (approved):
We should figure out a time to co-ordinate these alpenhorn renames.

PR #304, opened 2025-08-13
feat(QueryDatabase): Add a property to allow choosing HFB compression.

ketiltrout:
Setting the `hfb_compression` property to either true or false will also limit the search to absorber data.

Requires https://github.com/chime-experiment/ch_util/pull/98

### DIFF
--- a/ch_pipeline/core/dataquery.py
+++ b/ch_pipeline/core/dataquery.py
@@ -153,6 +153,9 @@ class QueryDatabase(task.MPILoggedTask):
         Reject time intervals that overlap with DataFlags of these types.
     return_intervals : bool (default: False)
         Return the full interval from the Finder. Otherwise only a list of file names.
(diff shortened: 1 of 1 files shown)


ljgray (approved):
This seems fine

ch_pipeline/core/dataquery.py
```diff
@@ -192,6 +195,8 @@ class QueryDatabase(task.MPILoggedTask):
 
     exclude_data_flag_types = config.Property(proptype=list, default=[])
 
+    hfb_compression = config.Property(proptye=bool, default=None)
```
ljgray, line 198:
```suggestion
    hfb_compression = config.Property(proptype=bool, default=None)
```

ljgray (approved):
Looks like a typo, this should be what's failing the test

PR #307, opened 2025-11-24
feat(processing.daily): Look in fir_smallfile for chimestack

ketiltrout:
I occasionally see in the alpenhorn log requests for chimestack transfers from nearline to online get cancelled because the file doesn't exist in nearline.  I think what's happening here, is the daily processing is looking for chimestack files which happen to be stored in `fir_smallfile` (because the files are too small).

So here's a change to the recall logic for the daily processing which will look in `fir_smallfile` for chimestack files if they can't be found in `fir_nearline`.


ljgray:
Although I seem to have failed spectacularly when re-exporting the new `fluxcat` repo into `ch_util`. I'll make a PR to fix that...

PR #318, opened 2026-03-10
Optionally provide pipeline status information as a json string

ljgray:
Will help to clean up automated access of these values from things like the pipeline monitor.


ketiltrout (approved):
This is great.

PR #320, opened 2026-04-28
feat(daily): Add fir_staging as a potential data source

ketiltrout:
Files will only be copied from staging if they don't exist elsewhere.


ljgray (approved):
Looks great! I'll try to remember to deploy in the next couple of days...

PR #325, opened 2026-07-13
feat(hfb): extract per-absorber ringmap cutouts from a catalogue

yukari-u:
Add new tasks that produce one ringmap cutout per absorber per sidereal day for a given absorber catalogue (confirmed and candidates).

The list of new tasks and container
- hfb.containers.AbsorberCatalogue
- hfb.io.LoadFilesForAbsorbers
- hfb.io.MakeAbsorberCatalogue
- hfb.analysis.SelectBeamsAroundSources
- hfb.analysis.ExtractAbsroberCutouts

### DIFF
--- a/ch_pipeline/hfb/analysis.py
+++ b/ch_pipeline/hfb/analysis.py
@@ -1,5 +1,6 @@
 """Tasks for HFB analysis."""
 
+import beam_model.formed as fm
 import numpy as np
 from beam_model.composite import FutureMostAccurateCompositeBeamModel
 from caput import config, mpiarray
@@ -8,9 +9,13 @@ from caput.containers import copy_datasets_filter, empty_like
 from caput.pipeline import tasklib
 from caput.util import mpitools
 from ch_util.hfbcat import HFBCatalog, get_doppler_shifted_freq
+from draco.analysis.sidereal import _search_nearest
+from draco.core import io
 from draco.util import tools
 from skyfield.positionlib import Angle
 from skyfield.starlib import Star
+
+from ch_pipeline.hfb.containers import HFBHighResRingMap
 
 from . import containers
 from .io import BeamSelectionMixin
@@ -954,6 +959,95 @@ class SelectBeam(BeamSelectionMixin, tasklib.base.ContainerTask):
         return newstream
 
 
+class SelectBeamsAroundSources(tasklib.base.ContainerTask):
+    """Select the beams around the sources recorded in the attributes.
+
+    The source declination and frequency are read from the container
+    attributes ('source_dec', 'source_freq'), which are set by
+    LoadFilesForAbsorbers.
+
+    Attributes
+    ----------
+    n_beams_ns : int
+        Number of NS beams to include on either side of the beam closest
+        to each source (i.e. 2 * n_beams_ns + 1 NS beams per source).
+        Default is 2.
+    beam_ew_include : list
+        List of East-West beam indices to include. Default is [0, 1, 2].
+    """
+
+    n_beams_ns = config.Property(proptype=int, default=2)
+    beam_ew_include = config.Property(proptype=list, default=[0, 1, 2])
+
+    def setup(self, manager):
+        """Load the beam model used to locate the sources' NS beams.
+
+        Parameters
+        ----------
+        manager :
+            An Observer object holding the geographic location of the telescope.
+        """
+        self.beam_mdl = fm.FFTFormedBeamModel()
+        self.observer = io.get_telescope(manager)
+        self.latitude = self.observer.latitude
+
+    def process(self, stream):
+        """Select the beams around the catalogue's sources.
+
+        Parameters
+        ----------
+        stream : containers.HFBData
+            Container with a beam axis and 'source_dec' / 'source_freq'
+            attributes.
+
+        Returns
+        -------
+        newstream : containers.HFBData
+            New container holding only the selected beams.
+        """
+        src_dec = np.atleast_1d(stream.attrs["source_dec"])
+        src_freq = np.atleast_1d(stream.attrs["source_freq"])
+
+        # Union of the NS-beam neighbourhoods of all absorbers in the catalogue.
+        ns_beams: set = set()
+        for dec, freq in zip(src_dec, src_freq):
+            angles = self.beam_mdl.get_beam_positions(np.arange(0, 256), [freq])
+            ns0 = int(_search_nearest(angles.T[1][0], dec - self.latitude))
+            ns_beams.update(
+                range(
+                    max(0, ns0 - self.n_beams_ns), min(256, ns0 + self.n_beams_ns + 1)
+                )
+            )
+
+        ns_sel = np.array(sorted(ns_beams))
+        ew_sel = np.array(self.beam_ew_include)
+
+        # Full beam indices, restricted to beams present in the stream
+        beam_sel = (ew_sel[:, np.newaxis] * 256 + ns_sel).flatten()
+        beam_sel = np.sort(beam_sel[np.isin(beam_sel, stream.beam)])
+
+        self.log.info(
+            f"Selecting {beam_sel.size} beams for {src_dec.size} absorbers "
+            f"({ns_sel.size} NS beams, EW {ew_sel.tolist()})."
+        )
+
+        # Create new container with the subset of beams
+        newstream = empty_like(stream, beam=beam_sel)
+
+        # Make sure all datasets are initialised
+        for dname in stream.datasets.keys():
+            if dname not in newstream.datasets:
+                newstream.add_dataset(dname)
+
+        # Find indices in current beam axis of selected subset of beams
+        selindex = np.flatnonzero(np.isin(stream.beam, beam_sel)).tolist()
+
+        # Copy over datasets
+        copy_datasets_filter(stream, newstream, "beam", selindex)
+
+        return newstream
+
+
 class HFBFlattenPFB(tasklib.base.ContainerTask):
     """Flatten HFB data using PFB deconvolution."""
 
@@ -1406,3 +1500,176 @@ class HFBMedianSubtraction(tasklib.base.ContainerTask):
         out.weight[:] = weight
 
         return out
+
+
+class ExtractAbsorberCutouts(tasklib.base.ContainerTask):
+    """Extract and save one cutout per absorber in the catalogue.
+
+    The frequency window is physical (MHz); the RA and el windows are in
+    pixels (samples) around the pixel nearest the source: e.g. the defaults
+    give the centre pixel +/- 30 RA pixels and +/- 1 el pixel (one beam
+    either side). The RA window wraps correctly around RA = 0/360; the el
+    window is clipped at the edges of the el axis.
+
+    Attributes
+    ----------
+    freq_window : float
+        Half-width of the frequency window to extract, in MHz.
+        Default is 0.2 (i.e. a 400 kHz wide cutout, ~131 high-res
+        channels).
+    n_ra : int
+        Number of RA pixels to include on either side of the pixel nearest
+        the source RA (i.e. 2 * n_ra + 1 pixels total). Default is 30.
+    n_el : int
+        Number of el pixels (beams) to include on either side of the pixel
+        nearest the source el (i.e. 2 * n_el + 1 pixels total, clipped at
+        the axis edges). Default is 1.
+    """
+
+    freq_window = config.Property(proptype=float, default=0.2)
+    n_ra = config.Property(proptype=int, default=30)
+    n_el = config.Property(proptype=int, default=1)
+
+    def setup(self, manager):
+        """Set the local observer's position.
+
+        Parameters
+        ----------
+        manager :
+            An Observer object holding the geographic location of the telescope.
+        """
+        self.observer = io.get_telescope(manager)
+        self.latitude = self.observer.latitude
+
+    def process(self, stream):
+        """Extract and save one cutout file per absorber in the catalogue.
+
+        Parameters
+        ----------
+        stream : containers.HFBHighResRingMap
+            High frequency resolution ringmap container.
+
+        Returns
+        -------
+        None
+            Files are saved directly via '_save_output'; nothing is passed
+            downstream.
+        """
+        from mpi4py import MPI
+
+        # Ensure ONLY the freq axis is distributed. After this, beam_ew, el
+        # and ra are complete on every rank.
+        stream.redistribute("freq")
+
+        names = np.atleast_1d(stream.attrs["source_names"])
+        src_ra = np.atleast_1d(stream.attrs["source_ra"])
+        src_dec = np.atleast_1d(stream.attrs["source_dec"])
+        src_freq = np.atleast_1d(stream.attrs["source_freq"])
+        lsd = stream.attrs.get("lsd", stream.attrs.get("csd", None))
+
+        freq = np.asarray(stream.index_map["freq"])
+        ra = np.asarray(stream.index_map["ra"])
+        el = np.asarray(stream.index_map["el"])
+        beam_ns = np.asarray(stream.index_map["beam_ns"])
+        nra = ra.size
+        nel = el.size
+
+        # Local data: (beam_ew, el, ra, freq_local)
+        hfb_local = stream.hfb[:].local_array
+        weight_local = stream.weight[:].local_array
+        freq_offset = stream.hfb[:].local_offset[3]
+        nfreq_local = hfb_local.shape[3]
+        nbeam_ew = hfb_local.shape[0]
+
+        comm = stream.comm
+        nsaved = 0
+
+        for iobj, name in enumerate(names):
+            name = str(name)
+
+            # Frequency: +/- freq_window MHz around the absorber frequency
+            fsel = np.flatnonzero(np.abs(freq - src_freq[iobj]) <= self.freq_window)
+            if fsel.size == 0:
+                self.log.warning(
+                    f"Absorber {name}: no data in frequency range. Skipping."
+                )
+                continue
+
+            # RA: nearest pixel +/- n_ra pixels, wrapping at 0/360
+            dra = (ra - src_ra[iobj] + 180.0) % 360.0 - 180.0
+            ira0 = int(np.argmin(np.abs(dra)))
+            rsel = (ira0 + np.arange(-self.n_ra, self.n_ra + 1)) % nra
+
+            # el: nearest pixel +/- n_el pixels, clipped at edges
+            el_src = np.sin(np.radians(src_dec[iobj] - self.latitude))
+            ie0 = int(np.argmin(np.abs(el - el_src)))
+            esel = np.arange(max(0, ie0 - self.n_el), min(nel, ie0 + self.n_el + 1))
+
+            # Extract this rank's freq chunk
+            shape = (nbeam_ew, esel.size, rsel.size, fsel.size)
+            data_cut = np.zeros(shape, dtype=hfb_local.dtype)
+            weight_cut = np.zeros(shape, dtype=weight_local.dtype)
+
+            # Convert global freq indices to this rank's local indices
+            fsel_local = fsel - freq_offset
+            valid = (fsel_local >= 0) & (fsel_local < nfreq_local)
+            fsel_local = fsel_local[valid]
+
+            # Destination positions along the cutout freq axis (0..nfsel-1)
+            fsel_out = np.flatnonzero(valid)
+
+            if fsel_local.size > 0:
+                ix = np.ix_(np.arange(nbeam_ew), esel, rsel, fsel_local)
+                data_cut[..., fsel_out] = hfb_local[ix]
+                weight_cut[..., fsel_out] = weight_local[ix]
+
+            # Combine freq chunks from all ranks
+            comm.Allreduce(MPI.IN_PLACE, data_cut, op=MPI.SUM)
+            comm.Allreduce(MPI.IN_PLACE, weight_cut, op=MPI.SUM)
+
+            if not weight_cut.any():
+                self.log.warning(f"Absorber {name}: all-zero weights. Skipping.")
+                continue
+
+            # Build and fill the output container
+            out = HFBHighResRingMap(
+                beam_ew=stream.index_map["beam_ew"],
+                beam_ns=beam_ns[esel],
+                el=el[esel],
+                ra=ra[rsel],
+                freq=freq[fsel],
+                attrs_from=stream,
+            )
+            out.redistribute("freq")
+
+            of = out.hfb[:].local_offset[3]
+            ns = out.hfb[:].local_array.shape[3]
+            out.hfb[:].local_array[:] = data_cut[..., of : of + ns]
+            out.weight[:].local_array[:] = weight_cut[..., of : of + ns]
+
+            # Replace catalogue-level array attrs with per-absorber scalars
+            for key in [
+                "source_names",
+                "source_ra",
+                "source_dec",
+                "source_freq",
+                "source_status",
+                "source_amplitude",
+            ]:
+                if key in out.attrs:
+                    del out.attrs[key]
+            out.attrs["source_name"] = name
+            out.attrs["source_ra"] = float(src_ra[iobj])
+            out.attrs["source_dec"] = float(src_dec[iobj])
+            out.attrs["source_freq"] = float(src_freq[iobj])
+
+            tag = name if lsd is None else f"{name}_lsd_{int(lsd)}"
+            out.attrs["tag"] = tag
+
+            outfile = self._save_output(out)
+            if outfile is not None:
+                self.log.info(f"Saved cutout for {name} -> {outfile}.")
+                nsaved += 1
+
+        self.log.info(f"Saved {nsaved}/{len(names)} cutouts.")
+        return
--- a/ch_pipeline/hfb/containers.py
+++ b/ch_pipeline/hfb/containers.py
@@ -1,5 +1,6 @@
 """HFB containers."""
 
+import re
 from functools import cached_property
 from typing import ClassVar
 
@@ -12,6 +13,7 @@ from draco.core.containers import (
     COMPRESSION_OPTS,
     DataWeightContainer,
     SiderealContainer,
+    SourceCatalog,
     TODContainer,
 )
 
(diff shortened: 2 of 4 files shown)


ch_pipeline/hfb/containers.py
```diff
@@ -783,3 +785,176 @@ def get_mask(
...
+    """A catalogue of absorbers (known and candidate).
+
+    Required per-entry values are: 'ra' (degrees), 'dec' (degrees), 'freq' (MHz),
+    and 'status'. The 'amplitude' is optional: when unknown it is stored as
+    NaN, and when provided it must be a finite float.
+
+    Status values
+    -------------
+    Allowed values are "confirmed", "false_positive", "control", and
+    candidates. A candidate records the S/N of its detection in the status
```
ljgray:
```suggestion
    "candidate". A candidate records the S/N of its detection in the status
```

yukari-u:
In response to feedback I received, I revised the structure. Here are the new tasks and containers to add:

- hfb.containers.HFBAbsorberCatalog
- hfb.io.MakeAbsorberCatalog
- hfb.io.LoadFilesForAbsorbers
- hfb.analysis.SelectBeamsAroundSources
- hfb.analysis.ExtractAbsroberCutouts
- hfb.io.UpdateAbsorberStacks

The absorber information used for this search is from: 
https://github.com/chime-experiment/ch_hfbcat/pull/1/changes

ch_pipeline/hfb/containers.py
```diff
@@ -515,6 +516,57 @@ class HFBHighResRingMap(HFBBeamRingMap, HFBHighResContainer):
...
+    _axes = ("csd",)
+
+    _dataset_spec: ClassVar = {
+        "hfb": {
+            "axes": ["csd", "beam_ew", "el", "ra", "freq"],
+            "dtype": np.float32,
+            "initialise": True,
+            "distributed": True,
+            "distributed_axis": "csd",
+            "compression": "gzip",
```
ljgray: Why `gzip` here instead of `bitshuffle`?

ch_pipeline/hfb/analysis.py
```diff
@@ -329,8 +355,35 @@ def process(self, stream):
         """
         from ch_ephem.coord import bmxy_to_hadec
```
ljgray:
```suggestion
```

This is already imported at the top of the module

ch_pipeline/hfb/analysis.py
```diff
@@ -329,8 +355,35 @@ def process(self, stream):
...
+
+        if self.mode == "wrap":
+            # "wrap" only makes sense for full RA coverage
+            dra = np.median(np.diff(ra))
+            if wraps or not np.isclose(ra.size * dra, 360.0, rtol=1e-3):
+                raise ValueError(
+                    "mode='wrap' requires an RA axis covering the full "
+                    "0-360 deg circle; got a partial RA window. Use "
+                    "mode='zero' or 'extrapolate' for cutouts."
+                )
```
ljgray: I'm not sure I understand this - if we had a window covering ra 355 - 5, for example, we would still want to wrap, even though we don't have full RA coverage.

ch_pipeline/hfb/analysis.py
```diff
@@ -351,22 +404,23 @@ def process(self, stream):
...
                 ) = _interpolation_linear(
                     x=ra_true,
                     y=data[iewb, insb, :, :],
                     w=weight[iewb, insb, :, :],
-                    xeval=stream.ra,
-                    mode="wrap",
-                    xperiod=360.0,
+                    xeval=ra.copy(),
+                    mode=self.mode,
+                    xperiod=360.0 if self.mode == "wrap" else None,
```
ljgray: `xperiod` is ignored if `mode != "wrap"`, so no need to check it here

ch_pipeline/hfb/analysis.py
```diff
@@ -351,22 +404,23 @@ def process(self, stream):
...
                 )
 
         # Create output container; add data and weights
-        out = containers.HFBHighResRingMap(copy_from=stream)
+        out = containers.HFBHighResRingMap(axes_from=stream, attrs_from=stream)
```
ljgray: This will not have the correct RA axis, since it copies the wrapped axis from the input container. If the desire is for the output stack to have the unwrapped axis, you'd need to explicitly pass `ra=ra`

ch_pipeline/hfb/analysis.py
```diff
@@ -954,6 +1010,97 @@ def process(self, stream):
...
+        newstream : containers.HFBData
+            New container holding only the selected beams.
+        """
+        src_dec = np.atleast_1d(stream.attrs["source_dec"])
+        src_freq = np.atleast_1d(stream.attrs["source_freq"])
+
+        # Union of the NS-beam neighbourhoods of all absorbers in the catalog.
+        ns_beams: set = set()
+        for dec, freq in zip(src_dec, src_freq):
+            angles = self.beam_mdl.get_beam_positions(np.arange(0, 256), [freq])
```
ljgray: `256` should come from somewhere (telescope nbeams?) instead of being hardcoded

ch_pipeline/hfb/analysis.py
```diff
@@ -954,6 +1010,97 @@ def process(self, stream):
...
+            ns0 = int(np.argmin(np.abs(angles[:, 0, 1] - (dec - self.latitude))))
+            #            angles = self.beam_mdl.get_beam_positions(np.arange(0, 256), [freq])
+            #            ns0 = int(_search_nearest(angles.T[1][0], dec - self.latitude))
```
ljgray:
```suggestion
```

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1553,243 @@ def process(self, stream):
...
+    dra_deg = config.Property(proptype=float, default=2.0)
+    n_coarse = config.Property(proptype=int, default=6)
+    n_el = config.Property(proptype=int, default=2)
+
+    def setup(self, manager):
+        """Set the local observer's position and load the beam model.
+
+        Parameters
+        ----------
+        manager :
```
ljgray:
```suggestion
        manager : Observer
```

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1553,243 @@ def process(self, stream):
...
+        freq = np.asarray(stream.index_map["freq"])
+        ra = np.asarray(stream.index_map["ra"])
+        el = np.asarray(stream.index_map["el"])
+        beam_ew = np.asarray(stream.index_map["beam_ew"])
+        nra = ra.size
+        nel = el.size
+        ra_step = float(np.median(np.diff(ra)))
+
+        # Number of HFB sub-frequencies per coarse channel.
+        nsubfreq = 128
```
ljgray: This probably shouldn't be hard-coded

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1553,243 @@ def process(self, stream):
...
+            raise RuntimeError(
+                f"el window ({n_el_win}) is wider than the el axis ({nel})."
+            )
+
+        # This rank's slice of the data, (beam_ew, el, ra, freq_local),
+        # and where that slice sits on the full freq axis.
+        hfb_local = stream.hfb[:].local_array
+        weight_local = stream.weight[:].local_array
+        freq_offset = stream.hfb[:].local_offset[3]
+        nfreq_local = hfb_local.shape[3]
```
ljgray: Shouldn't make assumptions about the exact index of a given axis (look for an axis name in the dataset attributes and determine the index from there)

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1553,243 @@ def process(self, stream):
...
+            fsel_local = fsel_local[valid]
+            fsel_out = np.flatnonzero(valid)
+
+            if fsel_local.size > 0:
+                ix = np.ix_(np.arange(nbeam_ew), esel, rsel, fsel_local)
+                data_cut[..., fsel_out] = hfb_local[ix]
+                weight_cut[..., fsel_out] = weight_local[ix]
+
+            # Combine freq chunks from all ranks
+            comm.Allreduce(MPI.IN_PLACE, data_cut, op=MPI.SUM)
```
ljgray: I might just be misunderstanding the desired operation here, but why is this a sum over ranks? Don't we want to concatenate frequencies instead of reducing that axis?

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1553,243 @@ def process(self, stream):
...
+        -------
+        cont : caput.memdata.MemDiskGroup
+            One group per absorber (cont[name]), each holding
+            beam_ew / el / ra / freq / hfb / weight datasets,
+            plus cont.attrs["csd"]. N
+        """
+        # Ensure only the freq axis is distributed. After this,
+        # beam_ew, el and ra are complete on every rank.
+        stream.redistribute("freq")
+        comm = stream.comm
```
ljgray: Use `self.comm` instead of extracting the comm from the container

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1553,243 @@ def process(self, stream):
...
+            fsel_out = np.flatnonzero(valid)
+
+            if fsel_local.size > 0:
+                ix = np.ix_(np.arange(nbeam_ew), esel, rsel, fsel_local)
+                data_cut[..., fsel_out] = hfb_local[ix]
+                weight_cut[..., fsel_out] = weight_local[ix]
+
+            # Combine freq chunks from all ranks
+            comm.Allreduce(MPI.IN_PLACE, data_cut, op=MPI.SUM)
+            comm.Allreduce(MPI.IN_PLACE, weight_cut, op=MPI.SUM)
```
ljgray:
```suggestion
            self.comm.Allreduce(MPI.IN_PLACE, weight_cut, op=MPI.SUM)
```

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+        self.log.info(f"Will iterate over {len(self.filegroups)} file groups.")
+
+        # Take the absorber parameters from the catalog container
+        if not isinstance(catalog, HFBAbsorberCatalog):
+            raise TypeError(f"Expected an HFBAbsorberCatalog, got {type(catalog)}.")
+        catalog.validate()
+
+        names = np.array([str(n) for n in catalog.index_map["object_id"]])
+        freqs = np.asarray(catalog["absorber"]["freq"][:])
+        freq_ax = np.linspace(800.0, 400.0, 1024, endpoint=False)
```
ljgray: This should be derived from the telescope class (`observer.freq_start, observer.freq_end, observer.num_freq`)

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+
+        time_range = filegroup.get("time_range", (None, None))
+
+        # Load each file into an HFBData container, applying the union
+        # frequency selection, then concatenate along the time axis
+        fgroup_containers = []
+        for fname in filegroup["files"]:
+            if not os.path.exists(fname):
+                raise RuntimeError(f"File does not exist: {fname}")
+            cont = HFBData.from_file(
```
ljgray: I would remove the unnecessary check here - if the file doesn't exist, `HFBData.from_file` will raise an error right away anyway,.

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+                fname,
+                distributed=self.distributed,
+                freq_sel=self.freq_sel,
+            )
+            fgroup_containers.append(cont)
+
+        tstream = (
+            _concatenate_time(fgroup_containers)
```
ljgray: What happens here if two absorbers have overlapping time indices? The behaviour of `concatenate_time` is pretty complicated, but I would think that either (a) all but the first dataset would get (partially?) ignored or (b) each incoming dataset would overwrite the overlapping part of the previous one that was inserted into the combined dataset

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+
+        Parameters
+        ----------
+        cont : caput.memdata.MemDiskGroup
+            The output of ExtractAbsorberCutouts: 'cont.attrs["csd"]' plus
+            one group per absorber (cont[name]), each holding
+            beam_ew / el / ra / freq / hfb / weight datasets.
+        """
+        comm = MPI.COMM_WORLD
+        rank0 = comm.rank == 0
```
ljgray: use `mpitools.rank0` and `self.comm`

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+                    error = f"{name}: {type(e).__name__}: {e}"
+                    break
+
+        # Stop the job cleanly if it failed.
+        error = comm.bcast(error, root=0)
+        if error is not None:
+            raise RuntimeError(f"UpdateAbsorberStacks failed on rank 0: {error}")
+
+        if rank0:
+            self.log.info(f"Merged {nsaved}/{len(names)} cutouts into their stacks.")
```
ljgray: Don't need to hide logging behind the `rank0` check

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+                self.log.warning("Nothing to merge (no cutouts this day).")
+            return
+
+        csd = int(cont.attrs["csd"])
+        names = list(cont.keys())
+
+        error = None
+        nsaved = 0
+
+        if rank0:
```
ljgray: Why only rank0?

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+            beam_ew, ra and freq must equal the stack file's axes while el only has to
+            have the same length.
+        data : dict
+            Dataset name -> array, e.g. {'hfb': ..., 'weight': ...}.
+        """
+        path = os.path.join(self.stack_dir, f"{name}.h5")
+        lockpath = path + ".lock"
+        tmp_path = os.path.join(self.stack_dir, f".{name}.{os.getpid()}.tmp.h5")
+
+        if not os.path.exists(path):
```
ljgray: Where does the file first get created?

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +454,476 @@ def setup(self, filelists):
...
+            raise RuntimeError(
+                f"{name}: stack file {path} does not exist. Create it before "
+                "running this task."
+            )
+
+        # The lock file is opened in append mode so it is created if missing.
+        with open(lockpath, "a") as lockfile:
```
ljgray: I would consider reworking this a bit. You can use `<container>.from_file(..., ondisk=True, mode="a")` to get a reference as an `h5py.File` object, without actually loading the data into memory. This also opens in `append` mode, which should protect the existing data (removing the need for the temporary file). `hdf5` files also have built-in locking, so opening it in this way also removes the need for the lock file.

You will also see a performance improvement, since you avoid having to load the entire stack every time you append to it.

ch_pipeline/hfb/analysis.py
yukari-u: I'm sorry I didn't intend to include this (HFBAlignEWBeams) in this PR. I'll revert it to the current version.

ch_pipeline/hfb/io.py
yukari-u: These are successive time chunks of a single acquisition, not per-absorber selections. While the frequency selection is the union over all absorbers and every file is read with the same freq_sel, there's no per-absorber time axis that could overlap, I think.

ch_pipeline/hfb/io.py
yukari-u: I add `create_absorber_stacks()` in `io.py`.  This reads the JSON target lists, computes each source's RA/el/freq window, and writes one zero-filled stack per source with a csd axis spanning given csd_range.

I think it's better not to add this to the UpdateAbsorberStacks task because several days can be processed concurrently and would race to create the same file. 

UpdateAbsorberStacks task can raise with a pointer to the function if a stack is missing.

yukari-u:
In response to feedback I received, I revised the structure. Here are the new tasks and containers to add:

hfb.containers.HFBAbsorberCatalog
hfb.io.MakeAbsorberCatalog
hfb.io.LoadFilesForAbsorbers
hfb.analysis.SelectBeamsAroundSources
hfb.analysis.ExtractAbsroberCutouts
hfb.io.UpdateAbsorberStacks

hfb.analysis.ra_window_pixels
hfb.io.create_absorber_stacks

The absorber information used for this search is from:
https://github.com/chime-experiment/ch_hfbcat/pull/1/changes

ljgray (approved):

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+
+    _done = False
+
+    def setup(self, manager):
+        """Set up the observer.
+
+        Parameters
+        ----------
+        manager :
+            An Observer object holding the geographic location of the teles>
```
rikvl: telescope

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+        if not files:
+            raise config.CaputConfigError(
+                "At least one JSON file must be given via 'json_files'."
+            )
+
+        # CHIME coarse channel centres
+        cfreq = np.linspace(
+            self.observer.freq_start,
+            self.observer.freq_end,
+            self.observer.num_freq,
```
rikvl: This is nice. But I don't understand where these properties of the `observer` (`freq_start`, etc.) come from. Could you explain?

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+                    float(entry["ra"]),
+                    float(entry["dec"]),
+                    float(entry["freq"]),
+                    str(entry.get("cal_src") or ""),
+                )
+            except (KeyError, TypeError, ValueError) as e:
+                raise RuntimeError(
+                    f"Entry {name} in {fname} is missing or has an invalid "
+                    f"required key: {e}."
+                ) from e
```
rikvl: This seems to duplicate the json-parsing code in `MakeAbsorberCatalog`

The two versions have diverged in how they handle `cal_src` on a duplicate `name`

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+            f"Read {len(files)} files and merged {nmerged} as duplicates; "
+            f"{len(entry_list)} unique absorbers and {len(cal_entries)} calibrator entries, "
+            f"totaling {len(names)}."
+        )
+
+        return cat
+
+
+class LoadFilesForAbsorbers(BaseLoadFiles):
+    """Load HFB data for the absorbers in a catalog.
```
rikvl:
```suggestion
class LoadFilesForCatalog(BaseLoadFiles):
    """Load HFB data for the sources in a catalog.
```
I would maybe rename this to something like `LoadFilesForCatalog` or `LoadFilesForAbsorberCatalog`, because the catalog can also contain calibration sources, and also because `LoadFiles` above is generally used to load files for absorbers

ch_pipeline/hfb/containers.py
```diff
@@ -515,6 +516,59 @@ class HFBHighResRingMap(HFBBeamRingMap, HFBHighResContainer):
     }
 
 
+class HFBHighResRingMapStack(HFBHighResRingMap):
```
rikvl: The `el` axis in this container has a different meaning than the `el` in its parent classes. If they will stay different [reference angles vs. more-or-less true sin(ZA)], I think they should have different names. We could maybe change the pre-existing `el` to `el_ref`,  for example (though not sure if this would break other things). Or you could rename the new one. The difference also needs to be documented in property docstrings.

ch_pipeline/hfb/io.py
```diff
@@ -1,19 +1,26 @@
...
+import json
+import logging
 import os
 from pathlib import Path
 
 import caput.astro.time as ctime
+import h5py
 import numpy as np
-from beam_model.formed import FFTFormedActualBeamModel
+from beam_model.formed import FFTFormedActualBeamModel, FFTFormedBeamModel
```
rikvl: Can you explain the difference between these beam models? If you deliberately use one vs the other, that's probably worth at least a comment in the code further down where it's used

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+                    self.log.exception(f"Absorber {name}: stack update failed: {e}")
+                    error = f"{name}: {type(e).__name__}: {e}"
+                    break
+
+        # Stop the job cleanly if it failed.
+        error = self.comm.bcast(error, root=0)
+        if error is not None:
+            raise RuntimeError(f"UpdateAbsorberStacks failed on rank 0: {error}")
+
+        self.log.info(f"Merged {nsaved}/{len(names)} cutouts into their stacks.")
```
rikvl: `nsaved` is only incremented on rank 0, so every other rank will keep logging "Merged 0/N"

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+            self.log.warning("Nothing to merge (no cutouts this day).")
+            return
+
+        csd = int(cont.attrs["csd"])
+        names = list(cont.keys())
+
+        error = None
+        nsaved = 0
+
+        if mpitools.rank0:
```
rikvl:
```suggestion
        if self.comm.rank == 0:
```
might be better

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+        self.log.info(
+            f"Read {len(files)} files and merged {nmerged} as duplicates; "
+            f"{len(entry_list)} unique absorbers and {len(cal_entries)} calibrator entries, "
+            f"totaling {len(names)}."
+        )
+
+        return cat
+
+
+class LoadFilesForAbsorbers(BaseLoadFiles):
```
rikvl:
```suggestion
class LoadFilesForCatalog(BaseLoadFiles):
```
is used in some docstrings and may be a better name. Or change those docstrings mentions of "LoadFilesForCatalog"

rikvl: Also, this inherits from `BaseLoadFiles`, but then bypasses the `setup()` from `BaseLoadFiles`. Maybe it should inherit directly from `io.BaseLoadFiles`?

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+        for f0 in freqs:
+            i0 = int(np.argmin(np.abs(freq_ax - f0)))
+            self._slices.append(
+                slice(max(0, i0 - self.n_coarse), min(nfreq, i0 + self.n_coarse + 1))
+            )
+
+        # Union of all absorber channel windows, as a sorted list of indices.
+        self.freq_sel = sorted(
+            {ch for sl in self._slices for ch in range(sl.start, sl.stop)}
+        )
```
rikvl: I think `self._slices` can just be `_slices`

ch_pipeline/hfb/io.py
```diff
@@ -448,3 +460,734 @@ def setup(self, filelists):
...
+                cal_seen.add((cs, e["channel"]))
+
+                if cs not in positions:
+                    raise RuntimeError(
+                        f"Calibration source {cs} (for {e['name']}) not "
+                        f"found in {self.combinedps_file}."
+                    )
+                cra, cdec = positions[cs]
+
+                # Name from the calibrator's own position: e.g. "041+79_3C_220.3".
```
rikvl:
```suggestion
                # Name from the calibrator's own position: e.g. "141+79_3C_220.3".
```
Should be consistent with the example in the class docstring above

ch_pipeline/hfb/analysis.py
```diff
@@ -1406,3 +1516,251 @@ def process(self, stream):
...
+            fsel_out = np.flatnonzero(valid)
+
+            if fsel_local.size > 0:
+                ix = np.ix_(np.arange(nbeam_ew), esel, rsel, fsel_local)
+                data_cut[..., fsel_out] = hfb_local[ix]
+                weight_cut[..., fsel_out] = weight_local[ix]
+
+            # Combine freq chunks from all ranks
+            self.comm.Allreduce(MPI.IN_PLACE, data_cut, op=MPI.SUM)
+            self.comm.Allreduce(MPI.IN_PLACE, weight_cut, op=MPI.SUM)
```
rikvl: Could you investigate if you can use `Gatherv` to rank 0 instead?

rikvl (commented):
Hey Yukari, this is really nice work! Thanks for doing this!

I left some in-line comments, and have a few more general questions:
1. There are a bunch of comments from Liam that are marked as "Resolved" but without a comment from you and some without code changes. I would echo many of those comments, but I don't know why they were resolved.
2. I wonder why `create_absorber_stacks` is a stand-alone function to be run by hand, and not a pipeline task, while `MakeAbsorberCatalog` _is_ a pipeline task.
3. Could you say how much this has been tested on real HFB data?

ch_pipeline/hfb/containers.py
yukari-u: This code is outdated. We no longer have this line.

ch_pipeline/hfb/containers.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Same as above. Removed

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/analysis.py
yukari-u: Addressed.

ch_pipeline/hfb/io.py
yukari-u: Addressed.

ch_pipeline/hfb/io.py
yukari-u: Addressed.

ch_pipeline/hfb/io.py
yukari-u: Liam agreed

ch_pipeline/hfb/io.py
yukari-u: Addressed.

ch_pipeline/hfb/io.py
yukari-u: Addressed.

ch_pipeline/hfb/io.py
yukari-u: Addressed.

ch_pipeline/hfb/io.py
yukari-u: It's in radiocosmology/driftscan/drift/core/telescope.py 
Line 212-214

ch_pipeline/hfb/io.py
yukari-u: Addressed

ch_pipeline/hfb/io.py
yukari-u: Addressed

ch_pipeline/hfb/io.py
yukari-u: Addressed.

ch_pipeline/hfb/containers.py
yukari-u: I started on this but backed out. Renaming `el` to `el_ref` in `HFBRingMapBase` propagates to `HFBHighResRingMapStack` through inheritance, and since `_axes` accumulates down the chain, giving the stack its own `el `would mean it carries both `el_ref `and `el`. Avoiding that means detaching the stack from `HFBHighResRingMap` and rebuilding it from `SiderealContainer` + `HFBHighResContainer`, duplicating a fair amount of code including the `beam_ns/beam_ew/ra` properties.

Given that it would also mean edits to `analysis.py`, `sidereal.py` and elsewhere, I decided against making this change for now. I've documented the difference in the docstring on `HFBHighResRingMapStack` instead.

ch_pipeline/hfb/io.py
yukari-u: Addressed

ch_pipeline/hfb/io.py
yukari-u: Addressed

ch_pipeline/hfb/analysis.py
yukari-u: After discussing with @rikvl in person, we conclude that this works fine at the current cutout sizes, but every rank holds a full copy of the cutout, so it may not scale. I've left it as is for now with a comment pointing at Gatherv as the possible alternative if that becomes a problem.

ch_pipeline/hfb/analysis.py
yukari-u: After discussing with @rikvl in person, we conclude that this works fine at the current cutout sizes, but every rank holds a full copy of the cutout, so it may not scale. I've left it as is for now with a comment pointing at Gatherv as the possible alternative if that becomes a problem.

ch_pipeline/hfb/io.py
yukari-u: Addressed

ch_pipeline/hfb/io.py
yukari-u: addressed

